PR #62, opened 2021-12-01
Initial implementation for JSON signatures

mgrand:
Create a mechanisms for serializing structs to signed JSON and verifying the signatures.


mgrand:
Before anyone asks why I have created three new crates, here are the reasons:
* Because signed structs will be needed by both Pyrsia nodes and their clients, we need a library for shared code. For this reason, I created `pyrsia_client_lib`.
* To make signed structs easy to use, I needed to define a procedural macro. Rust requires that procedural macros be in a separate crate by themselves. For his reason, I created `signed_struct`.
* We need to test the macros in `signed_struct`. Rust won't allow the test to be in the same create as the procedural macro. Putting it in `pyrsia_client_lib` makes Cargo complain about a circular dependency. I had to create a separate crate to put the test in. This is why I created `signed_struct_test`.

prince-chrismc (changes requested):
> here are the reasons

I feel strongly the location of you new crates should be approved upon. 

> their clients, we need a library for shared code

1. CLI should not be doing any "work" signing should be done in the node.
2. "clients" can pick the parts they need from the node.
    - Cargo is **extremely** flexible! You can nest crates in the project structure. [docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies)

> Rust requires that procedural macros be in a separate crate by themselves

Again the does not require it to be at the root of the repository. Cargo is flexible.
First example I found in mozilla github code base: https://github.com/mozilla/gecko-dev/blob/d36cf98aa85f24ceefd07521b3d16b9edd2abcb7/xpcom/rust/xpcom/Cargo.toml#L13

> Rust won't allow the test to be in the same create as the procedural macro

you can run integration tests. https://github.com/ferrous-systems/testing-proc-macros is a great example! You can also read https://doc.rust-lang.org/book/ch11-03-test-organization.html which I used in #80

prince-chrismc:
I suspect the failures https://github.com/pyrsia/pyrsia/runs/4463970462?check_suite_focus=true#step:4:1267 are because of incorrect changes to the `Cargo.lock`

https://github.com/pyrsia/pyrsia/pull/62/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL533-L538

mgrand:
@prince-chrismc You wrote
> I feel strongly the location of you new crates should be approved upon.
Ok. Please explain. What don't you like about the current location? If you have a better idea, where to you suggest that they go?

> 1. CLI should not be doing any "work" signing should be done in the node.
Again, what are your reasons for this?

Requiring signing to be done in the node opens a security risk. Do to that would require that we make private keys available to Pyrsia nodes. That is an unnecessary risk of compromising private keys. Because nodes are interacting with other nodes, we have to assume that they will be the target of hackers. Just the fact that private keys are available to nodes makes them more of a target. It will also make people question how secure we are.

Nodes support organizations will be particularly attractive targets. We should keep them out of nodes for this reason alone. 

Another reason to not do signing in nodes is that it would imply that nodes are accepting unsigned data. This opens up another security risk. One simple rule that I think we should follow is that if data is not signed it is not on a Persia node. Allowing unsigned data means that someone who is trying to attack a Pyrsia node can remain anonymous.

> 2. "clients" can pick the parts they need from the node.
>    
>    * Cargo is **extremely** flexible! You can nest crates in the project structure. [docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies)

Yes Cargo is flexible. I fail to understand what the point is that your are trying to make with that. I assume that you have some application of Cargo's flexibility in mind, but I do not know what.

> > Rust requires that procedural macros be in a separate crate by themselves
> 
> Again the does not require it to be at the root of the repository. Cargo is flexible. First example I found in mozilla github code base: https://github.com/mozilla/gecko-dev/blob/d36cf98aa85f24ceefd07521b3d16b9edd2abcb7/xpcom/rust/xpcom/Cargo.toml#L13

Agreed. That is why the procedural macro crate and its test crate are not in the root. They are under pyrsia-client-lib. I do not understand your point.

> you can run integration tests. https://github.com/ferrous-systems/testing-proc-macros is a great example! You can also read https://doc.rust-lang.org/book/ch11-03-test-organization.html which I used in #80

Again, I do not know what you are asking of me. I have not integration tests here. If you consider testing a proc macro an integration test rather than a unit test. I don't what to argue semantics.

Please clarify what it is you want of me. The only change you actually requested, that we do our signing in the nodes, I will not do.

mgrand:
@prince-chrismc 
> I suspect the failures https://github.com/pyrsia/pyrsia/runs/4463970462?check_suite_focus=true#step:4:1267 are because of incorrect changes to the `Cargo.lock`
I have made no manual changes to Cargo.lock. What I understand from your statement is that your believe that I have somehow done something that resulted in the Cargo.lock file being somehow wrong. I have no idea what to do about this. Any suggestions?  Is there a particular command that I can run in my local environment to replicate the problem?

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Getters are generated with the signature `fn field(&self) -> &type`.
+///
+/// Setters are generated as `fn field(&mut self, val: type)`. In addition to setting their field,
+/// the setters also call the `clear_json()` method provided by the `Signed` trait. This removes
+/// any JSON currently associated with the struct because it is no longer valid after the struct's
+/// field has been modified.
+#[proc_macro_attribute]
+pub fn signed_struct(args: TokenStream, input: TokenStream) -> TokenStream {
+    ////////////////////////////////////////////////////////////////////////////////////////////
+    // To debug this, uncomment the two println! macros and the tokens generated by the macros
```
efrisch, line 38: Rather than have commented out `println!`, can we switch to the `dbg!` macro?

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+                }
+                println!("generating output");
+                let output = quote! {
+                #[derive(serde::Serialize, serde::Deserialize)]
+                #[derive(signed_struct::SignedStructDerive)]
+                #ast
+                }
+                .into();
+                //println!("Output for signed_struct: {}", output);
+                output
```
efrisch, line 61: clippy thinks you should eliminate the let binding. I think that would go away with a `dbg!` here.

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    let mut field_names: HashSet<String> = HashSet::new();
+    for field in fields.named.iter() {
+        if field.vis != Visibility::Inherited {
+            return Err(syn::parse::Error::new(
+                field.span(),
+                "signed_struct requires all fields to be private",
+            ));
+        }
+        for id in field.ident.iter() {
+            field_names.insert(id.to_string());
```
efrisch, line 99: Since you are iterating the field_names here, couldn't you find the max `_jsonYYY` here rather than 102-110?

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    let ast = parse_macro_input!(input as DeriveInput);
+    match &ast.data {
+        syn::Data::Struct(ref struct_data) => {
+            match &struct_data.fields {
+                syn::Fields::Named(fields) => {
+                    match scan_fields(fields) {
+                        Ok((json_field_name, type_vec, field_ident_vec, setter_name_vec)) => {
+                            let struct_ident = &ast.ident;
+                            let output = quote! {
+                                impl<'π> ::pyrsia_client_lib::signed::Signed<'π> for #struct_ident<'π> {
```
efrisch, line 124: I think I remember you telling me you chose π because it was unlikely to appear in JSON; can we document that here. Or is my memory playing tricks on me... regardless; why π?

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+pub fn signed_struct_derive(input: TokenStream) -> TokenStream {
+    let ast = parse_macro_input!(input as DeriveInput);
+    match &ast.data {
+        syn::Data::Struct(ref struct_data) => {
+            match &struct_data.fields {
+                syn::Fields::Named(fields) => {
+                    match scan_fields(fields) {
+                        Ok((json_field_name, type_vec, field_ident_vec, setter_name_vec)) => {
+                            let struct_ident = &ast.ident;
+                            let output = quote! {
```
efrisch, line 123: This `quote!` macro is [beautiful](https://docs.rs/quote/latest/quote/).

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+
+                                    #(fn #field_ident_vec(&self)->&#type_vec{&self.#field_ident_vec}
+
+                                      fn #setter_name_vec(&mut self, value: #type_vec){self.clear_json(); self.#field_ident_vec = value}
+
+                                    )*
+                                }
+                            }
+                            .into();
+                            //println!("Output from signed_struct_derive: {}", output);
```
efrisch, line 151: `dbg!`?

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            )
+            .to_compile_error()
+            .into()
+        }
+    }
+}
+
+fn scan_fields(
+    fields: &FieldsNamed,
+) -> Result<
```
efrisch, line 180: Congrats! You've triggered a "very complex type used" warning from clippy. Reminds me of "insufficient politeness" from intercal. I wonder if there's a `[Please]` annotation.

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            "signed_struct_derive does not work with an empty struct",
+        )),
+    }
+}
+
+fn field_type_with_modified_lifetime(field: &Field) -> Type {
+    match field.ty {
+        syn::Type::Reference(ref t) => {
+            let mut ty = t.clone();
+            ty.lifetime = Some(Lifetime::new("'π", Span::call_site()));
```
efrisch, line 218: I feel like this 'π needs a comment. I find this surprising. What are we doing here? I can almost work it out from context.

pyrsia-client-lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+        _ => field.ty.clone(),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        let result = 2 + 2;
+        assert_eq!(result, 4);
```
efrisch, line 230: This does not feel a particularly high value unit test.

pyrsia-client-lib/src/signed.rs
```diff
@@ -0,0 +1,1240 @@
+//! This module defines structs and traits that are used to implement _signed structs_. A signed
+//! struct is a that has associated with it a JSON representation of the struct's contents that are
```
efrisch, line 2: "is a that"? A signed struct has a JSON representation with an embedded verifiable signature field.

pyrsia-client-lib/src/signed.rs
```diff
@@ -0,0 +1,1240 @@
...
+//! The first thing that you will do with this module is to use its `create_key_pair` function to
+//! create a key pair. You will use the private key in the key pair to sign structs. The public key
+//! is used to identify the signer.
+//! ```
+//! use pyrsia_client_lib::signed::{SignatureKeyPair, create_key_pair, JwsSignatureAlgorithms};
+//! let key_pair: SignatureKeyPair = create_key_pair(JwsSignatureAlgorithms::RS512).unwrap();
+//! ```
+//!
+//! The next thing to do is define some signed structs. Signed structs implement the `Signed` trait.
+//! However, it is not recommended that you directly implement the `Signed` trait directly. Instead,
```
efrisch, line 19: Redundant "directly" at end of sentence.

pyrsia-client-lib/src/signed.rs
```diff
@@ -0,0 +1,1240 @@
...
+//! you should annotate the struct like this <br>
+//! `   #[signed_struct]` <br>
+//! `   struct Foo<'a> {` <br>
+//! `       foo: String,` <br>
+//! `       bar: u32,` <br>
+//! `       zot: &'a str,` <br>
+//! `   }` <br>
+//!
+//! This annotation runs a macro that add some fields to support the Signed trait, implements the
```
efrisch, line 28: `Signed` trait

pyrsia-client-lib/src/signed.rs
```diff
@@ -0,0 +1,1240 @@
...
+//! signed trait, and generates getters and setters for the struct. There is not a full example of
+//! its use here to avoid Cargo complaining about a circular dependency. You can see a detailed
+//! example in the source for `signed_struct_test/lib.rs'. This is the best example of how to use
+//! this. You should read it.
```
efrisch, line 32: Redundant "this".

Cargo.toml
```diff
@@ -2,7 +2,8 @@
 
 members = [
     "pyrsia_node",
-    "pyrsia_cli"
+    "pyrsia_cli",
+    "pyrsia-client-lib"
```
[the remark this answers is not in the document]
mgrand: @efrisch pyrsia-client-lib is a lib crate for code that is needed for both Pyrsia nodes and clients. Clients will want to create and verify signatures. Nodes will want to verify sigantures. There are also some structs that will need to be shared, though they are not yet in there.

pyrsia-client-lib/signed_struct/src/lib.rs
mgrand: @efrisch The dbg! macro will still generate output. I have expanded the comment to further explain. Please let me know if it still does not make sense.

pyrsia-client-lib/signed_struct/src/lib.rs
mgrand: @efrisch I don't quite understand. I am trying _json0. If it clashes I try _json1, and then _json2 and so on.

pyrsia-client-lib/signed_struct/src/lib.rs
mgrand: @efrisch I documented this in the file comment. I wanted to choose something that would be unlikely to clash with another lifetime defined outside the macro. I choose π because in greek, pyrsia is πυρσός.

pyrsia-client-lib/signed_struct/src/lib.rs
mgrand: That would just generate more voluminous output that would be harder to understand.  See the new improved comment ad the beginning of the file.

pyrsia-client-lib/signed_struct/src/lib.rs
mgrand: @efrisch See the new improved comment at the end of the beginning of file comment.

pyrsia-client-lib/signed_struct/src/lib.rs
mgrand: @efrisch Thank you.

pyrsia-client-lib/src/signed.rs
```diff
@@ -0,0 +1,1271 @@
...
+        header,
+    )?;
+    let jws_string = String::from_utf8(jws)?;
+    let mut signed_json_buffer = before_string.clone();
+    if target.is_empty() {
+        // No existing signatures
+        signed_json_buffer.push('"');
+        signed_json_buffer.push_str(SIGNATURE_FIELD_NAME);
+        signed_json_buffer.push_str(r#"":[""#);
+        signed_json_buffer.push_str(jws_string.as_str());
```
MitaliBo, line 633: repetitive code can be move out of if-else

pyrsia-client-lib/src/signed.rs
```diff
@@ -0,0 +1,1271 @@
...
+            JsonPathElement::Field("__signature"),
+            JsonPathElement::Index(4),
+        ];
+        assert_eq!("path[field:\"__signature\",4]", path_to_str(&path))
+    }
+
+    #[test]
+    fn parse_json() -> Result<(), anyhow::Error> {
+        env_logger::try_init().unwrap_or_default();
+        let json = r#"{"boo":true,"number":234,"nul":null, "ob":{"a":123,"b":"str"}, "arr":[3, true, {"sig":"mund", "om":"ega"}, "asfd"] , "extra":"qwoeiru"}"#;
```
MitaliBo, line 1118: is it better to use json file for testing and move all below tests there?

prince-chrismc:
None of the new crates should be in the root of the repository. 

They should all be under the `pyrsia_node`

```
pyrsia_node
|_src/
|_signing
   |_src/
   |_signing_macros
      |_src/
      |_tests/
|_tests/
pyrsia_cli/
|_src/
```

pyrsia-client-lib/src/signed.rs
mgrand: Fixed.

Cargo.toml
```diff
@@ -2,7 +2,10 @@
 
 members = [
     "pyrsia_node",
-    "pyrsia_cli"
+    "pyrsia_cli",
+    "pyrsia-client-lib",
```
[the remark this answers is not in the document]
mgrand: fixed

pyrsia-client-lib/src/signed.rs
mgrand: Thank you. It is done.

pyrsia-client-lib/src/signed.rs
mgrand: @MitaliBo Putting the JSON in a file wouldn't help here. I added a comment to better explain what is going on.

prince-chrismc (approved):

pyrsia_node/pyrsia_client_lib/signed_struct/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+/// fields, you should use the getters and setters that the macro generates.
+///
+/// Getters are generated with the signature `fn field(&self) -> &type`.
+///
+/// Setters are generated as `fn field(&mut self, val: type)`. In addition to setting their field,
+/// the setters also call the `clear_json()` method provided by the `Signed` trait. This removes
+/// any JSON currently associated with the struct because it is no longer valid after the struct's
+/// field has been modified.
+///
+/// The generated impls and their methods use a lifetime named 'π. This greek letter was chosen
```
efrisch, line 36: 💯

pyrsia_node/pyrsia_client_lib/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#![allow(mixed_script_confusables)] // This is to allow structs created by a derive macro to have private fields that begin with the grek letter π
```
efrisch, line 1: Confusables! 🥇

pyrsia_node/pyrsia_client_lib/src/signed.rs
```diff
@@ -0,0 +1,1276 @@
...
+//! However, it is not recommended that you implement the `Signed` trait directly. Instead, you
+//! should annotate the struct like this <br>
+//! `   #[signed_struct]` <br>
+//! `   struct Foo<'a> {` <br>
+//! `       foo: String,` <br>
+//! `       bar: u32,` <br>
+//! `       zot: &'a str,` <br>
+//! `   }` <br>
+//!
+//! This annotation runs a macro that add some fields to support the Signed trait, implements the
```
efrisch, line 27: Missing \` around Signed.

pyrsia_node/pyrsia_client_lib/src/signed.rs
```diff
@@ -0,0 +1,1276 @@
...
+        } else if "RS384" == name {
+            Some(JwsSignatureAlgorithms::RS384)
+        } else {
+            None
+        }
+    }
+}
+
+// The default size for RSA keys
+const DEFAULT_RSA_KEY_SIZE: u32 = 8192;
```
efrisch, line 118: 💪

pyrsia_node/pyrsia_client_lib/src/signed.rs
```diff
@@ -0,0 +1,1276 @@
...
+        month,
+        day,
+        time.hour,
+        time.minute,
+        time.second,
+        time.millisecond,
+        time.tz_offset_hours,
+        time.tz_offset_minutes
+    );
+    println!("{}", formatted_date_time);
```
efrisch, line 259: `dbg!` ?

pyrsia_node/pyrsia_client_lib/src/signed.rs
```diff
@@ -0,0 +1,1276 @@
...
+    .unwrap();
+    OffsetDateTime::parse(&formatted_date_time, &format).unwrap()
+}
+
+/// An instance of this struct is created to hold a key pair. The struct can be serialized to save
+/// the key pair for later use.
+///
+/// This struct has the public and private keys as separate values in anticipation that quantum-
+/// resistant signature algorithms will require this.
+#[derive(Deserialize, Serialize)]
```
efrisch, line 272: `Debug`, `Clone` and `Copy` might also be good practice. https://doc.rust-lang.org/rust-by-example/trait/clone.html

efrisch (approved):
👍

PR #80, opened 2021-12-07
Move integration tests out of unit tests

prince-chrismc:
Following/learning https://doc.rust-lang.org/book/ch11-03-test-organization.html plus some Stack overflow links in the commits which gave more details.


mgrand (dismissed):

prince-chrismc:
I update the test coverage to run over the entire workspace as well

pyrsia_node/utilities/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
betarelease, line 1: Can we choose a folder name that is more descriptive of this folder should contain?
`utilities` is very generic and everything that does not easily fit in other modules gets thrown in a it can become a mess. If this is a library for doing more specific things we should use that name instead.

prince-chrismc: How about `test_helpers`?

mgrand (approved):

pyrsia_node/utilities/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+use std::io::BufRead;
+
+// Reads the first line from a BufRead
+pub fn first_line<R>(mut rdr: R) -> String
+where
```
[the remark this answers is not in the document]
prince-chrismc: Excellent question! I was following a tutorial and a6581d7 was sharing this code with unit and integration tests.

To do this, you need to have a crate that's a dev dependency

pyrsia_node/tests/integration_get.rs
```diff
@@ -0,0 +1,38 @@
+extern crate pyrsia_node;
```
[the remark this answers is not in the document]
prince-chrismc: If we have more tests we can rename the file

tests/test_code_coverage.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+docker build --no-cache -t code_coverage:2.0 -f ./tests/code_coverage.Dockerfile .
```
efrisch, line 3: Why bump the version on our ephemeral code_coverage docker image? We explicitly remove it on the next line.

efrisch (dismissed):
Assuming it merges clean 👍

tests/test_code_coverage.sh
tiainen: The image itself is not removed, only the container that is started from that image.

pyrsia_node/readme.md
```diff
@@ -9,8 +9,8 @@ The daemon running everything.
 
 ### Generating Test Coverage Report
 
-1. `cd pyrsia`
-2. `sh ./test_code_coverage.sh`
+1. From the root directory of the repository
+2. Run `sh ./test_code_coverage.sh`
```
tiainen, line 13: The script has moved to the `tests` directory:
```suggestion
2. Run `sh ./tests/test_code_coverage.sh`
```

tests/test_code_coverage.sh
prince-chrismc: Good catch!

prince-chrismc: I changed the inner structure and the command so that a "api break" so I bumped the version

prince-chrismc:
testing CLA ... RIP me

MitaliBo (approved):

mgrand (approved):

PR #83, opened 2021-12-08
Upload code coverage results

prince-chrismc:
My fork testing https://github.com/prince-chrismc/pyrsia-sandbox/pull/1


prince-chrismc:
You can see it here 

https://codecov.io/github/pyrsia/pyrsia

johanvos:
I ran this locally, but it consumes a few GB's diskspace and it takes a long time.
I believe this is very useful, but maybe some parameters can be tweaked so that it doesn't fill your disk everytime you run `sh test_code_coverage.sh` ?
If this is not intended to be run on a developers machine, this is less of an issue.

prince-chrismc:
That's an excellent point. 🤔 

Tarpaulin is massive, but there are precompiled binaries that might work on your system (depending on rust version and libc) which are fairly small. That could be a solution for most.

Given the local output is annoying to look at I personally don't think I would use it locally. The results are posted on PR which pretty graphs.

I expect we will eventually write a better docker image to make a smaller image and something  reusable but that's out of scope for me

mgrand (approved):
This looks like it will be helpful.

MitaliBo (approved):

PR #84, opened 2021-12-08
peer discovery

MitaliBo:


prince-chrismc:
> does this replace #84 they sound similar

that PR actually contains the IPC named pipes which kind of was not working well with p2p. So i am gonna close that one

_Originally posted by @MitaliBo in https://github.com/pyrsia/pyrsia/issues/111#issuecomment-995259069_

Issue #39, opened 2021-11-02
Create a document store layer above key value store for looking up metadata.


mgrand:
# Background
Our architecture calls for a document store to hold metadata and blockchain pages. We are planning to use an library named [UnQLite](https://docs.rs/unqlite/1.5.0/unqlite/) to implement our document store. UnQLite does not provide quite the functionality that we need, so we need to implement a layer on top of it that will be an adapter.

We want to be able to:
* Create named persistent collections of JSON objects.
* Configure a named collection have named indexes. Each index will used named fields of JSON object as key values.
* Add a JSON object to a named collection, which will automatically add it to all of the named indexes for the collection. Conditionally including JSON objects in an index is out of scope.
* Given a named collection, a named index and the relevant key value(s) retrieve the corresponding JSON object. Searches for multiple JSON objects are out of scope.

Update and deletion of JSON objects is out of scope for this task.

What UnQLite gives us out of the box is:
* A key-value store that works with keys and values that are arrays of bytes.
* A scripting language named Jx9 that provides part of a document store on top of the key-value store. It allows us to store records whose value is a given JSON object and whose key is a generated `u64` value.
* The ability to access the JSON objects in a collection, but no indexing capability.
* The ability to use the key associated with a JSON object to retrieve it.

Not the there is no built-in indexing mechanism. We will need to build an indexing mechanism that uses supplementary
key-value collection as indexes to map index keys to JSON record keys. 

# The Task
This task consists of these steps:
* Define a Rust struct named `json_collection_descr` that describes a JSON collection and its named indexes. It must describe the field names that contain the key value(s) for each named index.

* Define a Rust trait that defines methods to do the following things without in any way depending on using UnQLite for an implementation. We may decide on a different implementation in the future. 
  - Create an empty collection of JSON objects given a name and a `json_collection_descr`.
  - Add a JSON object to a named collection and update the named indexes as an atomic transaction. Either the JSON object is added to the collection and it is included in the indexes or none of that happens. <br>For now, we will treat all fields used for indexes as required. If they are missing or null we just reject the JSON object.
  - Given the name of a collection, the name of one of its indexes and a key value(s) retrieve the corresponding JSON object.

* Create an implementation of the rust trait that uses UnQLite.

# Completion Criteria
This task will be considered complete when:
* The code for this is in a new git repo named pyrsia-document-store.
* The implementation of the trait passes a set of unit tests.
* The implementation is available in a Rust crate.
[...]

PR #88, opened 2021-12-09
Add json support to `document_store`

### LINKED ISSUE #87: Add json support for documentstore
The current documentstore is text-oriented and does some very basic parsing.
As part of this issue, we want to replace the text protocol with standard json.

tiainen:
Fixes #87 

Storing a document requires the document to be a valid JSON string.
Fetching a document requires the search filter to be a valid JSON string.

### DIFF
--- a/pyrsia_node/src/document_store/document_store.rs
+++ b/pyrsia_node/src/document_store/document_store.rs
@@ -139,9 +139,7 @@ impl DocumentStore {
         db_name: &str,
         document: &str,
     ) -> Result<(), Box<dyn std::error::Error>> {
-        let db = self.dbs.get(db_name).ok_or(DocumentStoreError::Custom {
-            message: format!("No DocumentStore found for name {}", db_name),
-        })?;
+        let db = self.get_db(db_name)?;
 
         let json_document = serde_json::from_str::<Value>(document)?;
 
@@ -177,29 +175,22 @@ impl DocumentStore {
 
     // Fetch a document from the database with the provided
     // name that has an index that maps with the provided
-    // JSON filter
+    // indices
     pub fn fetch(
         &mut self,
         db_name: &str,
-        filter: &String,
+        indices: Vec<(&str, &str)>,
     ) -> Result<Option<String>, Box<dyn std::error::Error>> {
-        let db = self.dbs.get(db_name).ok_or(DocumentStoreError::Custom {
-            message: format!("No DocumentStore found for name {}", db_name),
-        })?;
+        let db = self.get_db(db_name)?;
 
-        let json_filter = serde_json::from_str::<Value>(filter)?;
-        let conditions = json_filter.as_object().ok_or(DocumentStoreError::Custom {
-            message: format!("Provided filter must be a JSON object"),
-        })?;
-
-        for (key, value) in conditions.iter() {
+        for (key, value) in indices.iter() {
             debug!("key = {}, value = {}", key, value);
             let mut index_name = String::from(db_name);
             index_name.push('_');
             index_name.push_str(key);
             debug!("this belongs to index db {}", index_name);
             if let Some(field_db) = self.fields.get(&index_name) {
-                if let Ok(key) = field_db.kv_fetch(value.as_str().unwrap()) {
+                if let Ok(key) = field_db.kv_fetch(value) {
                     debug!("Exists! key = {:?}", key);
                     if let Ok(raw_document) = db.kv_fetch(key) {
                         debug!("raw document = {:?}", raw_document);
@@ -212,6 +203,14 @@ impl DocumentStore {
         }
 
         Ok(None)
+    }
+
+    fn get_db(&self, db_name: &str) -> Result<&UnQLite, Box<dyn std::error::Error>> {
+        self.dbs.get(db_name).ok_or_else(|| {
+            From::from(DocumentStoreError::Custom {
+                message: format!("No DocumentStore found for name {}", db_name),
+            })
+        })
     }
 }
 
@@ -307,12 +306,8 @@ mod tests {
         doc_store
             .store(name, &doc.to_string())
             .expect("empty value");
-        let flt = json!({
-            "mostSignificantField": "msf1",
-            "leastSignificantField": "12"
-        });
         let res: String = doc_store
-            .fetch(name, &flt.to_string())
+            .fetch(name, vec![("mostSignificantField", "msf1")])
             .expect("Should have fetched without error.")
             .expect("Should have been found!");
         info!("Got result: {}", res);
@@ -341,43 +336,11 @@ mod tests {
         doc_store
             .store(name, &doc.to_string())
             .expect("empty value");
-        let flt = json!({
-            "mostSignificantField": "msf2",
-            "leastSignificantField": "12"
-        });
-        let res = doc_store.fetch(name, &flt.to_string());
+        let res = doc_store.fetch(name, vec![("mostSignificantField", "msf2")]);
         debug!("Got result: {:?}", res);
         assert_eq!(
             true,
             res.expect("Should have fetched without error.").is_none()
         );
-    }
-
-    #[test]
-    fn test_fetch_invalid_json() {
-        let tmp_dir = tempfile::tempdir().unwrap();
-        let path = tmp_dir.path().join("test_fetch_invalid_json");
-        let name = path.to_str().unwrap();
-        let mut doc_store = DocumentStore::new();
-        let n1 = "mostSignificantField".to_string();
-        let t1 = "string".to_string();
-        let n2 = "leastSignificantField".to_string();
-        let t2 = "string".to_string();
-        let i1 = Index::new(n1, t1, true);
-        let i2 = Index::new(n2, t2, false);
-        let idx = vec![i1, i2];
-        doc_store.create_db(name, idx);
-        let doc = json!({
-            "foo": "bar",
-            "mostSignificantField": "msf1",
-            "leastSignificantField": "12"
-        });
-        doc_store
-            .store(name, &doc.to_string())
-            .expect("empty value");
-        let flt = "{\"mostSignificantField\": \"msf2\"";
-        doc_store
-            .fetch(name, &String::from(flt))
-            .expect_err("Should not store invalid json.");
     }
 }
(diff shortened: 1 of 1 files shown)


pyrsia_node/src/document_store/document_store.rs
```diff
@@ -112,70 +107,87 @@ impl DocumentStore {
...
-                        return Some(String::from_utf8(raw_document).unwrap());
+        if let Ok(json_filter) = serde_json::from_str::<Value>(filter) {
+            if let Some(conditions) = json_filter.as_object() {
+                for (key, value) in conditions.iter() {
+                    debug!("key = {}, value = {}", key, value);
+                    let mut index_name = String::from(db_name);
+                    index_name.push('_');
+                    index_name.push_str(key);
+                    debug!("this belongs to index db {}", index_name);
+                    if let Some(field_db) = self.fields.get(&index_name) {
```
prince-chrismc, line 171: Personally I hate nested `if`s like this but rust gives use little choice? best I could find https://users.rust-lang.org/t/how-to-write-not-if-let/7251/3 but I am not sure thats relevant

prince-chrismc (approved):
LGTM

mgrand:
I am just starting to read through this and will have detailed feedback tomorrow.  For now I have some observations not related to the new code.

You are putting the indexes for a database each in a separate UnQLite database. This is going to be a problem for ensuring consistency. We need inserts into the database to atomically add a new record and update the indexes. It must not be possible for some to be updated and some not. 

UnQLite provides a mechanism for ensuring this through its transaction API. You tell it that you are starting a new transaction. When you are done inserting everything, you tell it you are committing the transaction. If there is a problem, you tell it to roll back the transaction and then none of the inserts happened. If the database crashes in the middle of a transaction, on restart it will recognize the a transaction was in progress and roll it back.

There is one limitation to all this. Transactions are implemented for individual key-value stores. If the indexes are each in their own key-value store, the transaction mechanism won't help. To be able to use the transaction mechanism, the indexes will need to be in the same key-value store as the data they index. 

To accommodate this, I suggest that you prefix each key in a key-value store with a byte that indicates that the key-value pair is one of three types:
* Data – If the first byte of the key indicates that it is a data key, then the rest of the key is a unique u64 value. The value part of the KV pair contains the record's JSON data.

* Index – If the first byte of a key indicates that it is an index key, then it is followed by a u16 that indicates which index it is part of.
   This does imply that a database will have no more that 64k indices. The u16 is followed by the values of the indexed fields. 
   The key of the record that contains the indexed values is in in the value part of the KV pair. 

* Metadata – There will only be one record of this type. The type byte is the entire key
  The value part of this will contain a JSON object that contains the name of the database, the names of the indexes and the fields that each index is based on.

johanvos:
> For now I have some observations not related to the new code.

Those are good observations, but not relevant to this issue (#87). We do not want to fix all problems in a single big PR.

pyrsia_node/src/document_store/document_store.rs
tiainen: I've improved the error handling, which reduces the levels of indentation.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -112,70 +132,86 @@ impl DocumentStore {
...
-    pub fn store(&mut self, db_name: &str, document: &str) -> Result<(), Error> {
-        let key = Key::new();
+    pub fn store(
+        &mut self,
+        db_name: &str,
+        document: &str,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let db = self.dbs.get(db_name).ok_or(DocumentStoreError::Custom {
+            message: format!("No DocumentStore found for name {}", db_name),
+        })?;
```
mgrand, line 144: You will be doing this check for a the existence of the DB for every API call, so please refactor this into a separate function.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -32,8 +41,18 @@ impl Index {
...
+        }
+    }
+}
+
 impl Key {
-    pub fn new() -> Key {
+    fn new() -> Key {
         let mut v: Vec<u8> = Vec::new();
         for _i in [0..7] {
             let rr = rand::random::<u8>();
```
mgrand, line 58: We will need to have some initial seeding of the random number generator to be sure that we always start with a different seed. There is some guidance on how to do this in https://rust-random.github.io/book/guide-seeding.html

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -112,70 +132,86 @@ impl DocumentStore {
...
-        // split filter in a number of conditions
-        let conditions: Vec<&str> = json.split(',').collect();
-        for condition in &conditions {
-            debug!("condition = {}", condition);
-            let (key, value) = split_key_value(condition);
+    // JSON filter
+    pub fn fetch(
+        &mut self,
+        db_name: &str,
+        filter: &String,
```
mgrand, line 184: Looking at how this has to be implemented, I think it would be better to change the method signature to include the name of the index to use. That will allow this method to be simpler and more performant.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -248,13 +299,24 @@ mod tests {
...
+            .store(name, &doc.to_string())
+            .expect("empty value");
+        let flt = json!({
+            "mostSignificantField": "msf1",
+            "leastSignificantField": "12"
+        });
+        let res: String = doc_store
+            .fetch(name, &flt.to_string())
+            .expect("Should have fetched without error.")
+            .expect("Should have been found!");
```
mgrand, line 317: The first expect will panic if there is an error, so the second expect will never be reached if there is an error.

mgrand (approved):

pyrsia_node/src/document_store/document_store.rs
tiainen: I agree, while working on this I came to the same conclusion.

pyrsia_node/src/document_store/document_store.rs
tiainen: I was actually already refactoring this for #89 to use [unqlite::Util::random_num](https://docs.rs/unqlite/1.4.0/unqlite/trait.Util.html#tymethod.random_num). I'm not sure if that is succifient or that a different RNG should be used.

pyrsia_node/src/document_store/document_store.rs
tiainen: Indeed, that's the idea of expect. It panics in case an error occurs. This code also panics in case there was no error but no result was found.

pyrsia_node/src/document_store/document_store.rs
mgrand: If you believe that will be suitable then that's fine.

pyrsia_node/src/document_store/document_store.rs
tiainen: Refactored into a separate function.

pyrsia_node/src/document_store/document_store.rs
tiainen: I've changed it so that `fetch` takes a Vec of tuple strings.

prince-chrismc (approved):
Much easier to read!

prince-chrismc:
@johanvos can you link Mark's outstanding comments to the other related issue(s)? We can break them in sprint planning?

#39 maybe? This way we can recall what's to do :)

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -177,29 +175,22 @@ impl DocumentStore {
...
-        })?;
+        let db = self.get_db(db_name)?;
 
-        let json_filter = serde_json::from_str::<Value>(filter)?;
-        let conditions = json_filter.as_object().ok_or(DocumentStoreError::Custom {
-            message: format!("Provided filter must be a JSON object"),
-        })?;
-
-        for (key, value) in conditions.iter() {
+        for (key, value) in indices.iter() {
```
mgrand, line 186: I don't quite understand this logic. I assume that you are looping over the search filters to construct the multiple values that may be in a compound key used by an index that is based on multiple fields of a JSON object. However the logic that you have does not look like it is doing that. It looks like it is treating the values as alternatives to be tried.

Some indexes will be for records that are identified by a combination of multiple values. For example, a PackageVersion will be identified the the combination of a packageType, a namespace id and a package name.

If you are treating this values as alternative search targets that is not what we need. If I am misunderstanding what you are doing, please make it clearer by putting comments in the code that explain what the methods are doing.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -177,29 +175,22 @@ impl DocumentStore {
 
     // Fetch a document from the database with the provided
     // name that has an index that maps with the provided
-    // JSON filter
+    // indices
     pub fn fetch(
         &mut self,
         db_name: &str,
-        filter: &String,
+        indices: Vec<(&str, &str)>,
```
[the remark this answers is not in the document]
tiainen: The current implementation allows fetching documents by indexed fields only. Each key that is passed in with the key/value parameter should reflect an index that was defined when creating the document store.

What you are saying is that you want to pass in a list of index names that should be used during lookup. Can you explain this in more detail in #39?

pyrsia_node/src/document_store/document_store.rs
tiainen: Compound indexes are not yet implemented. We can tackle this in a separate issue. For how the fetch implementation currently works, [see my comment above](https://github.com/pyrsia/pyrsia/pull/88#discussion_r767154340).

tiainen:
> @johanvos can you link Mark's outstanding comments to the other related issue(s)? We can break them in sprint planning?
> 
> #39 maybe? This way we can recall what's to do :)

@prince-chrismc I've created issue #94 to tackle this.

pyrsia_node/src/document_store/document_store.rs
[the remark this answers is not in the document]
mgrand: I have added a very detailed explanation in #39 .

johanvos (approved):

prince-chrismc:
> I've created issue #94 to tackle this.

Awesome! This is a great increment

PR #91, opened 2021-12-10
Move signed struct test to integration section

prince-chrismc:
part of the default cargo layout

@mgrand
- How do you feel about this? (purely moving files)
- Is there anything "path specific" in the comments I should also change?


Cargo.toml
```diff
@@ -4,8 +4,7 @@ members = [
     "pyrsia_node",
     "pyrsia_cli",
     "pyrsia_node/pyrsia_client_lib",
-    "pyrsia_node/pyrsia_client_lib/signed_struct",
-    "pyrsia_node/pyrsia_client_lib/signed_struct_test"
+    "pyrsia_node/pyrsia_client_lib/signed_struct"
```
mgrand, line 7: Why did you remove "pyrsia_node/pyrsia_client_lib/signed_struct_test"?  

signed_struct needs to be tested like everything else. Its test is in a separate crate because Rust does not allow the test for a proc macro to be in the same crate. A proc macro crate may contain ONLY proc macros.

I don't really care where it is in the repo, but we need to be able to run it as part of Cargo test.

pyrsia_node/pyrsia_client_lib/signed_struct/Cargo.toml
```diff
@@ -13,3 +13,8 @@ quote = "1.0"
 proc-macro2 = "1.0"
 pyrsia_client_lib = { path = ".." }
 syn = { version = "1.0", features = ["full"] }
+
+
+[dev-dependencies]
```
mgrand, line 18: Why did you add these dependencies?  It was fine without them.

Cargo.toml
prince-chrismc: > Why did you remove "pyrsia_node/pyrsia_client_lib/signed_struct_test"?

I did not remove anything. I simply **moved** them to a different location that cargo natively understands.
It was just a rename of the files.

> we need to be able to run it as part of Cargo test

They are. https://github.com/pyrsia/pyrsia/runs/4487064726?check_suite_focus=true#step:5:449

Cargo is incredibly flexible and it support this exact use case. The way you implemented it was not required.

My structuring the code like this:

```
signed_structs
|_src/
|_tests/ # Formally ../signed_struct_tests
```

the tests are run from out side of the root of the crate allowing them to access the procedural macro code (exactly like you had it).

➡️  This difference is this is the "official" way of structuring the code according to the rust handbook and cargo documentation.

pyrsia_node/pyrsia_client_lib/signed_struct/Cargo.toml
prince-chrismc: Because they were previously required for `signed_struct_tests` which was only a "development" crate.

By moving the tests to the integration section cargo is better able to manage them.

Cargo.toml
prince-chrismc: I renamed 

```diff
-"pyrsia_node/pyrsia_client_lib/signed_struct_test"
+"pyrsia_node/pyrsia_client_lib/signed_struct/tests"
```

mgrand: The addition of 
+"pyrsia_node/pyrsia_client_lib/signed_struct/tests"
does not appear in the diff.  I am looking a changes from **all commits**.

mgrand: Also, have you verified that the test runs?  Its name is `test_generated_methods`

prince-chrismc: > does not appear in the diff

https://github.com/pyrsia/pyrsia/pull/91/files#diff-292c3c34476b763266a0f1d27113ef08c53a8ca7dec0d541656a9caa2dfb3561

> have you verified that the test runs? Its name is `test_generated_methods`

https://github.com/pyrsia/pyrsia/runs/4487386582?check_suite_focus=true#step:5:446

mgrand (approved):

efrisch (approved):
👍

Issue #89, opened 2021-12-09
Store and fetch multiple documents


tiainen:
At the moment, the implementation of the document store has the following limitations:

1. store the first index that is detected in a document
2. storing a document overwrites the previous index for identical index values
3. fetch a single document for a given index

The implementation should be updated to support the following cases:

1. store each index that is defined for a document store
2. store multiple documents that have an identical index value
3. return a list of matching documents

mgrand:
@tiainen The behavior you describe is not what we want. An attempt to insert a record that would duplicate an index record needs to be rejected. I am currently in the process of documenting this in https://github.com/pyrsia/pyrsia/issues/126.

Consider this use case: we want to update the collection of versions in a packages. The outcome of doing that should be that there is one package record, but its `versions` field contains an additional version string.

Please close this and you can do #126 after #125.

PR #101, opened 2021-12-14
Atomically store documents to the `document_store`

### LINKED ISSUE #94: Atomically store documents to the document_store
Based on the https://github.com/pyrsia/pyrsia/pull/88#issuecomment-990436112 by @mgrand, storing a document and its indexes should happen atomically.

The current UnQLite implementation stores all necessary data spread over different key/value stores. Using UnQLite's transaction support, we can ensure atomic storage. This does however require the data to be stored in a single key/value store, because transactions are only spanning one individual store.

Below, I'll quote the suggestion of @mgrand on how to tackle this issue:

> To accommodate this, I suggest that you prefix each key in a key-value store with a byte that indicates that the key-value pair is one of three types:
> - Data – If the first byte of the key indicates that it is a data key, then the rest of the key is a unique u64 value. The value part of the KV pair contains the record's JSON data.
> - Index – If the first byte of a key indicates that it is an index key, then it is followed by a u16 that indicates which index it is part of. This does imply that a database will have no more that 64k indices. The u16 is followed by the values of the indexed fields. The key of the record that contains the indexed values is in in the value part of the KV pair.
> - Metadata – There will only be one record of this type. The type byte is the entire key. The value part of this will contain a JSON object that contains the name of the database, the names of the indexes and the fields that each index is based on.

tiainen:
This PR fixes #94. 

A document store is now always associated with exactly one UnQLite database. The database can hold three types of data:

1. catalog: contains the name and the index specification of the document store
2. data: represents a document in the document store
3. index: represents the index value of a document

This PR allows to store a document in a document store with zero or more indexes. Each index can be associated with multiple field names.

This PR further introduces a method to get a document store via its name. And I also took the time to start adding rustdoc.

This PR does **not** fix the following issues:

1. store and fetch multiple documents that have an identical index value, this is described in #89 
2. apply the order that is specified on an index


mgrand:
It should not be possible to create a document collection with no indexes. We currently have no way to access such a collection. I do not see a use case for supporting such a collection.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
-            debug!("Creating Field DB with name {}", field_name);
-            let field_db = UnQLite::create(format!("{}.unql", &field_name));
-            self.fields.insert(field_name, field_db);
+        let mut rng = rand::thread_rng();
 
-            all_indices.push(index);
+        let mut doc_store_indexes: Vec<(u16, IndexSpec)> = vec![];
+        for index in indexes {
+            doc_store_indexes.push((
+                rng.gen(),
```
mgrand, line 201: Is there a reason that you are generating random numbers for the index identifiers?  There is no reason I see no to just assign consecutive numbers. That changes the risk of index identifier collisions from 1 in 64K to zero. 

I do not think we will be dropping indexes, so it would be acceptable to have the collection of indexes be a Vec<IndexSpec> rather than a Vec<(u16,IndexSpec)>. That way the Index identifier would just be an index on the vector.

tiainen:
Following the same logic, an index should also define at least one field?

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+        let doc_store = bincode::deserialize(&raw_doc_store)?;
+
+        Ok(doc_store)
+    }
+
+    /// Store the provided JSON document in the document store.
+    pub fn store(&self, document: &str) -> Result<(), Box<dyn std::error::Error>> {
+        let db = DocumentStore::get_db(&self.name);
 
         let json_document = serde_json::from_str::<Value>(document)?;
```
mgrand, line 234: If the value is valid JSON, but not an object, you give a helpful message that explains the problem. You should probably add something equally helpful for this case where it is not valid JSON.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+
+        let mut raw_data_key = bincode::serialize(&DataKey::new(rng.gen()))?;
+        loop {
+            if !db.kv_contains(&raw_data_key) {
+                db.kv_store(&raw_data_key, document)
+                    .map_err(DocumentStoreError::UnQLite)?;
+                debug!("Document stored!");
+                break;
+            }
+            raw_data_key = bincode::serialize(&DataKey::new(rng.gen()))?;
```
mgrand, line 255: If you move this to the top of the loop, you can get rid of the statement before the loop.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+    /// // store a document
+    /// let document = json!({"field": "value"});
+    /// doc_store.store(&document.to_string());
+    ///
+    /// let mut filter = HashMap::new();
+    /// filter.insert("field", "value");
+    /// if let Ok(documents) = doc_store.fetch("index", filter) {
+    ///     println!("Found documents on index 'index': {}", documents);
+    /// }
+    /// ```
```
mgrand, line 311: Please add something to clean up the file that this test will create. I am reading this sequentially, so if this is taken care of later in the file, ignore this comment.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+            let mut values: Vec<String> = vec![];
+
+            for field_name in &index_to_use.1.field_names {
+                if let Some(value) = filter.get(&field_name as &str) {
+                    values.push(value.to_string());
                 } else {
-                    debug!("Does not exist!");
+                    return Err(From::from(DocumentStoreError::Custom {
+                        message: format!(
+                            "Filter is missing required index key: {}.{}",
```
mgrand, line 327: Might be clearer if formatted as `"Filter is missing value for field {} required by index {}", field_name, index_name`

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+                            index_name, field_name
+                        ),
+                    }));
+                }
+            }
+
+            let db = DocumentStore::open_db(&self.name);
```
mgrand, line 334: This is OK for now, but I expect that we will be more performant if we keep the underlying UnQLite collection open.

If you are concerned about the number of open files, there are currently 14 different collections implied by the metadata model. The blockchain may add another one or two. 

If you are concerned about the thread safety of these calls, that is something we can deal with later. I suspect that UnQLite is not thread safe, but please check that. If it is not thread safe, then we will need to ensure that operations on each instance of DocumentStore are single threaded.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+                        db.rollback().map_err(DocumentStoreError::UnQLite)?;
+                        return Err(From::from(DocumentStoreError::UnQLite(e)));
+                    }
+                }
+                Err(e) => {
+                    db.rollback().map_err(DocumentStoreError::UnQLite)?;
+                    return Err(e);
                 }
             }
         }
```
mgrand, line 279: Moving this for loop to a separate function would make this more readable.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+            if let Ok(raw_data_key) = db.kv_fetch(raw_index_key) {
+                if let Ok(raw_data) = db.kv_fetch(&raw_data_key) {
+                    debug!("Found raw data: {:?}", raw_data);
+                    results.push(String::from_utf8(raw_data).unwrap());
+                } else {
+                    let data_key: DataKey = bincode::deserialize(&raw_data_key)?;
+                    error!(
+                        "DocumentStore failed to find Document with key {} on index {}.",
+                        data_key.number, index_name
+                    );
```
mgrand, line 347: It looks like this error does not get returned. It just falls down to<br>
return Ok

mgrand (approved):

pyrsia_node/src/document_store/document_store.rs
tiainen: This is not an error, but an error message being printed. The explanation for the fetch function in #39 mentions to log an error and return None() in case no document exists for a key found via an index.

pyrsia_node/src/document_store/document_store.rs
tiainen: That sounds reasonable. I would like to keep the index stored in the catalog, just in case we would need drop index later on. It's not such a big value after all and simplifies having to determine the index key all the time.

pyrsia_node/src/document_store/document_store.rs
tiainen: This part is only a piece of example code to include with the documentation. And we currently have not made deleting document stores a feature.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -1,77 +1,112 @@
...
+    field_names: Vec<String>,
+    direction: IndexOrder,
 }
 
-pub struct Index {
+/// The document store is able to store and fetch documents
+/// with a list of predefined indexes.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct DocumentStore {
     name: String,
```
prince-chrismc, line 36: Why does it need a name? is the variable name not enough?

The tests just use the path 😕 

https://github.com/pyrsia/pyrsia/pull/101/files#diff-c82424dfff1e622cd18a61d68ebdce08a82fcd6f5ba4b118b600ebfd8fbed8b2R386-R388

prince-chrismc (commented):
This got super complicated 😖 I apologize if I did not follow all the past conversations but I have a ton of questions. 

> What was wrong with https://github.com/pyrsia/pyrsia/pull/101/files#diff-2c5d9cbb9033cbd429ad804a903684afdac3c2f5cb96bd0c474b56f0793521fbL52-L53>

Questions:
- Why is `new` private? (it's only called by `create` that I found) 
  - Can these be one function?
- Why doesnt `create` just return the "new" `DocumentStore` as a result
  - What purpose outside of this does `get` try to solve?

All these methods could be static and replace `self` with `name` and we'd have a singleton 😆 

Why doesnt the document store own whatever https://github.com/pyrsia/pyrsia/pull/101/files#diff-c82424dfff1e622cd18a61d68ebdce08a82fcd6f5ba4b118b600ebfd8fbed8b2R178 outputs? Instead of name which means it has to recreate it every time we need access?

***

Sorry for my ill-informed questions I just want to follow the development

prince-chrismc:
I am looking at https://github.com/pyrsia/pyrsia/blob/63ce66b5a9a78756bbe0928f7aec5dd33d626bf4/pyrsia_node/src/document_store/document_store.rs#L292

Can we schedule a zoom so you can explain what it's doing?

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -1,77 +1,112 @@
...
+    Desc,
+}
 
-pub struct DocumentStore {
-    dbs: HashMap<String, UnQLite>,
-    fields: HashMap<String, UnQLite>,
-    indices: HashMap<String, Vec<Index>>,
+/// The definition of an index in the document store.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct IndexSpec {
```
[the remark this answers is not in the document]
mgrand: @prince-chrismc It is called a specification because it specifies the name of the index and the names of the JSON fields that the index will be based on.

pyrsia_node/src/document_store/document_store.rs
mgrand: @prince-chrismc We will have multiple collections of JSON object for things in the metadata mode like packages and package versions.  To distinguish between these collections, we give them names. If we want to find a package, we will look in the package collection.

Each instance of DocumentStore will be associated with one collection. The name field is the name of the collection it is associated with.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -1,77 +1,112 @@
...
-#[derive(Debug)]
-pub struct Key {
-    elements: Vec<u8>,
+const KEYTYPE_CATALOG: u8 = 0b00000001;
+const KEYTYPE_DATA: u8 = 0b00000010;
+const KEYTYPE_INDEX: u8 = 0b00000011;
+
+/// A key that is associated with the metadata of the
+/// document store. It is identified by [KEYTYPE_CATALOG].
+#[derive(Debug, Deserialize, Serialize)]
```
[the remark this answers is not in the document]
mgrand: It is uncertain what the future of DocumentStore will be. My assumption is that we want to communicate our intentions as clearly as possible to promote reuse and maintainability.

pyrsia_node/src/document_store/document_store.rs
mgrand: Yes. Definitely keep the index in the catalog.

pyrsia_node/src/document_store/document_store.rs
mgrand: When tests are run the code blocks in documentation comments are treated as tests. This code will be compiled and run when you run tests.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
-        assert_eq!(2 + 2, 4);
+    fn test_create() {
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let path = tmp_dir.path().join("test_create");
+        let name = path.to_str().unwrap();
+
+        let result = DocumentStore::create(name, vec![]).expect("should not result in error");
+        assert_eq!(result, ());
+
+        let doc_store = DocumentStore::get(name).expect("should not result in error");
```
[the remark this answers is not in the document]
mgrand: @prince-chrismc The factory pattern is useful when you want to vary the type of object being created by configuration or run time parameters. We are not trying to vary anything here, so there is no need.

mgrand:
@prince-chrismc Some of your questions about the examples may be explained if you read a detailed comment I wrote towards the end of issue https://github.com/pyrsia/pyrsia/issues/39.

@tiainen Though you have copied pieces of my documentation into the code, it would be helpful to have the whole thing as 
```
//! File level Documentation comments
```
at the beginning of `document_store.rs`. It will answer questions that others looking at this for the first time will have.

pyrsia_node/src/document_store/document_store.rs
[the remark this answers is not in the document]
prince-chrismc: Ahhh! Okay that make sense... Specification is a precise requirement, and I can imagine that fitting here but I think something with less connotations/associations might be less confusing

How about?

 ```suggestion
pub struct IndexArrangement {
```

Since this is trying describe how they are interpreted to calculate the index (hopefully I understand this more)

File: pyrsia_node/src/document_store/document_store.rs (hunk 9 above)
Comment (prince-chrismc) [reply pyrsia_node/src/document_store/document_store.rs:36]: Multiple collections is fine. The "name" is as always set to "path" which started my confusion.
This "name" is the file path (maybe) of the unqlite data.

It seems a touch erroneous to save the "name" when it only need to (re)create that unqlite data access. Which occurs in most method of this type.

From a blindly reading code that I am not familiar with, It's hard to see the reason behind this.

File: pyrsia_node/src/document_store/document_store.rs (hunk 12 above)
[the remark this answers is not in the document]
Comment (prince-chrismc) [reply pyrsia_node/src/document_store/document_store.rs:391]: I apologize my comment was unclear, is this supposed to be a factory pattern? That's was this section of code looks like but as you point out 

> there is no need

**Comment by prince-chrismc** [comment]
Okay, thanks for linking the relevant conversation. It's hard to imagine needing something this complex so early but there's no harm.

***

My other API question(s) remains, I think the following could be change to better follows rust idioms

```diff
-   DocumentStore::create("sample", indexes);
-   let doc_store = DocumentStore::get("sample");
+   let doc_store = DocumentStore::new("sample", indexes);
```

pyrsia_node/src/document_store/document_store.rs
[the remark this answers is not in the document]
mgrand: @prince-chrismc Arrangement makes a lot less sense to me. My suggestion is that it is not worth our time to discuss the name of a struct.

I notice that you have not yet approved the PR. I have not seen you mention anything in particular that you want Joeri to do before you approve. If there is something that you think he needs to do before you can approve, then please tell him. If there is nothing that you need him to do and you have provided all of your feedback, then you should approve.

pyrsia_node/src/document_store/document_store.rs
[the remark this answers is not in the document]
mgrand: If it were an instance of the Factory pattern, the details of the struct creation would be hidden behind a separate factory struct whose only purpose for existence would be to create DocumentStore structs.

mgrand:
@prince-chrismc If you still have questions, please schedule a time for us to talk. Joeri is implementing my design, so I can explain what the code is doing.

pyrsia_node/src/main.rs
```diff
@@ -49,8 +49,10 @@ async fn main() {
     pretty_env_logger::init();
 
     // create the connection to the documentStore.
-    let doc_store = DocumentStore::new();
+    DocumentStore::create("document_store", vec![]).expect("Failed to create DocumentStore");
+    let doc_store = DocumentStore::get("document_store").unwrap();
```
prince-chrismc, line 53: ## Zoom Summary Notes

The break through was the `create` is for the persistent storage of the document store!

Sometimes we want to "reload" the existing one hence the role of a `get` method.

- [ ] naming of "create" could be more verbose.

We dont know if there will be issues (for local development) so we will table any discussion for when problem a rise.

***

Some brain storming ideas that lead us to a consensus (maybe)




```suggestion
let doc_store = DocumentStore::create_if_no_exists_presistent_storage("name", vec![], /*append or overwrite*/) // Fails when called twice
```

```suggestion
let doc_store = DocumentStore::get("old name")
```

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
+    /// filter is a map of which the keys represent each field name
+    /// of the specified index. The values represent the associated
+    /// values of the keys in the documents to fetch.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// // create a document store
+    /// let indexes = vec![
+    ///     IndexSpec::new("index", vec!["field".to_string()], IndexOrder::Asc),
```
prince-chrismc, line 297: ## Zoom Summary Notes

- IndexOrder is not required for MVP
- Implies more work to support it (currently just a place holder)

- [ ] We can drop this from the API
- [ ] Leave and just skip the implementation

- "rebuilding document store" --> Versioning schematics? future work idea --> reorganizing

pyrsia_node/src/document_store/document_store.rs
prince-chrismc: ## Zoom Summary Notes

This name is the "label" to _reload_ (via `get`) and existing **persistent** document store. 

My break through was the instance of document store does not own a the file on disk and many instance can refer to the same file.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -109,238 +174,441 @@ impl DocumentStore {
...
-                if let Ok(key) = field_db.kv_fetch(value) {
-                    debug!("Exists! key = {:?}", key);
-                    if let Ok(raw_document) = db.kv_fetch(key) {
-                        debug!("raw document = {:?}", raw_document);
-                        return Ok(Some(String::from_utf8(raw_document).unwrap()));
-                    }
+        &self,
+        index_name: &str,
+        filter: HashMap<&str, &str>,
+    ) -> Result<Vec<String>, Box<dyn std::error::Error>> {
```
mgrand, line 316: @tiainen This should return Option<String> rather than Vec<String>. There will not be more than one record with the same index values. Any attempt to create such a record would update the existing record if we didn't first check for its existence. Also, when you do a looking in the KV store like this, you only match a single record. 

There will be a different method to return multiple records.

pyrsia_node/src/document_store/document_store.rs
tiainen: I noticed this isn't true when building a binary project. But I'll make sure it gets cleaned up.

pyrsia_node/src/document_store/document_store.rs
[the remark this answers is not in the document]
tiainen: Replaced rustdoc comments with regular comments.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -153,192 +223,319 @@ impl DataKey {
...
 
-    fn get_db(db_name: &str) -> UnQLite {
-        UnQLite::create(format!("{}.unql", &db_name))
+    fn get_data_store(name: &str) -> UnQLite {
+        UnQLite::create(format!("{}.unql", name))
     }
 
-    fn open_db(db_name: &str) -> UnQLite {
-        UnQLite::open_mmap(format!("{}.unql", &db_name))
+    fn open_data_store(name: &str) -> UnQLite {
```
mgrand, line 261: There should not be a need for this if we are caching the `UnQLite` in the `DocumentStore`.

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -153,192 +223,319 @@ impl DataKey {
...
+                        &data_store,
+                        &json_document,
+                        &raw_data_key,
+                        index.0,
+                        &index.1,
+                    ) {
+                        data_store.rollback().map_err(DocumentStoreError::UnQLite)?;
+                        return Err(e);
+                    }
+                }
```
mgrand, line 404: moving this to a separate function would make it more readable.

mgrand (approved):
This is close. I had a few comments that will make this more robust and performant. You can address them in this PR or just write separate issues for them.

pyrsia_node/Cargo.toml
```diff
@@ -38,10 +39,12 @@ stringreader = "0.1.1"
...
 strum_macros = "0.22.0"
 tempfile = "3.2.0"
+test-log = "0.2.8"
 tokio = { version = "1", features = [ "macros", "rt-multi-thread", "io-std" ] }
 unqlite = "1.5.0"
 uuid = { version = "0.8.2", features = [ "v4" ] }
 warp = { version = "0.3.1", default-features = false }
 
 [dev-dependencies]
+env_logger = "*"
```
prince-chrismc, line 49: Seems a little out of place since this was already in the cargo tree.

Can you confirm it's required?

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -24,17 +58,17 @@ pub enum IndexOrder {
...
 }
 
 /// The document store is able to store and fetch documents
 /// with a list of predefined indexes.
 #[derive(Debug, Deserialize, Serialize)]
 pub struct DocumentStore {
-    name: String,
-    indexes: Vec<(u16, IndexSpec)>,
+    pub name: String,
+    pub indexes: Vec<(u16, IndexSpec)>,
```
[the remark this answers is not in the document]
prince-chrismc: > I suggest that you add an `unqilite` field to this struct, so that we can cache the result of an open operation in the struct.

I agree! https://github.com/pyrsia/pyrsia/pull/101#pullrequestreview-832079659

prince-chrismc (approved):
Good enough! Lots a great future work notes, it be nice to save them in an issue for "after demo"

Cant wait to see this integrated into with the rest of our living demo 😉

Issue #95, opened 2021-12-13
Create an initial implementation of the metadata manager


mgrand:
This task is to:
* Create and organize for use as signed structs the structs and enums used to represent metadata for:
  * package types
  * namespaces
  * package
  * package version
  * artifact
  
* Design the initial metadata manager API to support
  * create namespace
  * get namespace
  * get namespaces by package type
  * create package
  * get package
  * update package 
  * get packages by namespace
  * update package
  * create package version

* Implement the API to call the document store or a mock of it.

mgrand:
Created a draft pull request for the first bullet.

prince-chrismc:
> Created a draft pull request for the first bullet.

Can you please link the related pull request(s)?

betarelease:
Marking as outdated. And closing.

PR #102, opened 2021-12-14
Refactor model and fix bug in signed_struct

mgrand:
Refactor model to be complete for the demo and use<br>
#[signed_struct]

Also fixed bug in 
#[signed_struct] 
that prevented it from working correctly for structs that did not specify a lifetime parameter.

For issue #95


pyrsia_node/Cargo.toml
```diff
@@ -22,11 +22,13 @@ hyper = { version = "0.14", features = ["full"] }
 hyper-tls = "0.5.0"
 log = "0.4.14"
 pretty_env_logger = "0.4.0"
+pyrsia_client_lib = { path = "pyrsia_client_lib" }
```
prince-chrismc, line 25: You should need to include both.

Can we try removing this one?

```suggestion
```

pyrsia_node/src/main.rs
```diff
@@ -26,7 +26,7 @@ extern crate warp;
 #[macro_use]
 extern crate lazy_static;
 //local module imports
-mod artifact_manager;
+pub mod artifact_manager;
```
prince-chrismc, line 29: Main is not exposed 🤔 

```suggestion
mod artifact_manager;
```

prince-chrismc (commented):
Clippy also left a few good suggestions! (In the new code)

prince-chrismc:
I'd like to improve the PR title... 

This does two things
1. Fixes #95
2. Bug in signing without lifetime

Is that correct?

pyrsia_node/Cargo.toml
mgrand: Both are needed. They are separate crates. The fact that the source for one is stored under the other does not change that. Modules nest. Crate are all top-level.  

My own feeling about nesting the source directories is that it leads to this confusion. I would like to have a source organization that matches what we are building. That was the reason for the original organization that you objected to. 

However, I did try deleting this and got
```
  --> pyrsia_node/src/node_manager/model/artifact.rs:16:1
   |
16 | extern crate pyrsia_client_lib;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
```

pyrsia_node/src/main.rs
mgrand: I do not understand this comment. 

The reason that I made this change is that without this I was getting MANY warning messages about functions in the artifact_manager not being called. By making artifact_manager public, the compiler is no longer sure that the functions are not being called and so the warnings stop.

pyrsia_node/src/node_manager/model/artifact.rs
```diff
@@ -13,21 +13,26 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 */
-
+extern crate pyrsia_client_lib;
 extern crate serde_json;
+
 use super::HashAlgorithm;
+use pyrsia_client_lib::signed::Signed;
```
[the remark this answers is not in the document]
mgrand: Again, I am not sure what your point is. `extern crate` declaration are not generally required. They are a good practice because it announces to cargo that this module will be using the crate. If the crate is missing from the Cargo.toml, you get a more helpful error message.

mgrand:
> I'd like to improve the PR title...
> 
> This does two things
> 
> 1. Fixes [Create an initial implementation of the metadata manager #95](https://github.com/pyrsia/pyrsia/issues/95)
> 2. Bug in signing without lifetime

No that is not correct. I gave it the title I gave it to describe the main point of the PR. 

It was suggested last week that we should have small PRs. for that reason I divided 95 up into three main bullet points. It is the first of three installments to accomplish 95. The main point of this is to refactor the model to work as signed structs. Fixing the bug was incidental. I discovered the bug in #[signed_struct] when I was doing the refactoring and so needed to fix it.

The title accurately reflects the intent of the PR when I created it as a draft. 

With a suggestion so general as "improving the PR title" I don't really know what you want.

mgrand:
@prince-chrismc Is there anything you need me to do before you approve?

pyrsia_node/src/main.rs
MitaliBo: if there are more than 1 pub fn defined in module than we do not need to define pub mod

pyrsia_node/src/node_manager/model/identity.rs
```diff
@@ -0,0 +1,34 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+extern crate pyrsia_client_lib;
+extern crate serde;
+extern crate serde_json;
+
+use pyrsia_client_lib::signed::{JwsSignatureAlgorithms, Signed};
+use serde::{Deserialize, Serialize};
```
MitaliBo, line 21: please remove unused imports, its good to have zero warnings

MitaliBo (dismissed):
Minor comments nut not a blocker

pyrsia_node/src/main.rs
mgrand: @MitaliBo I'm sorry but I don't understand. What is the connection between whether a module if public and how many public functions it has?

prince-chrismc: Nothing in a file called `main.rs` can be exposed so adding "pub" here does nothing.

Please remove it.

prince-chrismc:
The title will be the merge commit message. it's extremely vague. It should not have your name in it.
Please add a more descriptive title that other can understand. It's not enough only you understand your intention.

Here's a great read to learn about good commit messages https://cbea.ms/git-commit/

pyrsia_node/src/main.rs
MitaliBo: @mgrand you mentioned about getting warnings in artifact_manager for functions not being called, so my comment was for that. Also it would have been better if artifact_manager followed the same protocol as other modules in the tree are following  , is to define a mod file and declare that module public in there and then just call in main, so that from main all the visibility part is hidden

pyrsia_node/Cargo.toml
prince-chrismc: I just noticed there's a circular reference.... No wonder this looks wrong.

https://github.com/pyrsia/pyrsia/blob/0ebe6fade49a4967f16d686afed7341af2a22626/pyrsia_node/pyrsia_client_lib/signed_struct/Cargo.toml#L14

prince-chrismc: https://github.com/pyrsia/pyrsia/issues/110

pyrsia_node/src/node_manager/model/artifact.rs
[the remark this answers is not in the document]
prince-chrismc: Related to https://github.com/pyrsia/pyrsia/issues/110

pyrsia_node/src/node_manager/model/identity.rs
mgrand: Done

pyrsia_node/src/main.rs
mgrand: Done.

prince-chrismc (approved):

mgrand:
@MitaliBo Please re-review PR 102. You approved it previously, but I made some changes and need your approval again.

MitaliBo (approved):

PR #115, opened 2021-12-16
Add apt-get packaging and upload to repo

sbtaylor15:
The new steps will create the .deb file and then upload it to a apt repo that is hosted in the pyrsia/website repo as Github Pages.


sbtaylor15:
One outstanding issue is that the upload steps are running for both the macOS and Ubuntu when it should only run for Ubuntu.

sbtaylor15:
Found issue with macOS persisting files at the same time at ubuntu.  closing until resolved.

PR #116, opened 2021-12-16
Add apt-get packaging and upload to repo

### LINKED ISSUE #79: `apt-get install pyrsia` on Ubuntu LTS
Package for pyrsia is installed and it starts the daemon for pyrsia

https://github.com/pyrsia/pyrsia/issues/59

sbtaylor15:
The action will now create the .deb file and upload it to the apt repo that is hosted on the pyrsia/website repo as GitHub Pages. The package creation/repo upload only happens when the Action is running on the pyrsia/pyrsia repo and not a fork.

Closes #79  #59


.github/workflows/rust.yml
```diff
@@ -29,6 +29,84 @@ jobs:
...
+      if: ${{ github.repository_owner == 'pyrsia' && runner.os == 'Linux' }}   
+      with:
+        name: pyrsia
+        path: target/debug/pyrsia
+
+# save the DEBIAN control files to be used with the deb packager
+    - uses: actions/upload-artifact@v2
+      if: ${{ github.repository_owner == 'pyrsia' && runner.os == 'Linux' }}    
+      with:
+        name: DEBIAN
```
prince-chrismc, line 50: I see this is checked into the repo does it need to be uploaded?

prince-chrismc: Maybe with a checkout in the next job?

.github/workflows/rust.yml
```diff
@@ -29,6 +29,84 @@ jobs:
...
+            echo "Installed-Size: $(echo "$(stat -c%s pyrsia/usr/local/bin/pyrsia) + $(stat -c%s pyrsia/usr/local/bin/pyrsia_node)" | bc)" >> pyrsia/DEBIAN/control; \
+            sed -i.bak "/Version:/s/\$/-$GITHUB_RUN_NUMBER/" pyrsia/DEBIAN/control
+            cat pyrsia/DEBIAN/control; \
+            dpkg-deb --build pyrsia; \
+            ls -la pyrsia.deb;
+        shell: bash
+
+# add the prysia.deb file to the apt repo which is part of
+# the github website repo
+      - uses: sbtaylor15/apt-repo-action@v2.0.4
```
prince-chrismc, line 94: Is it possible to specify the commit message? It would be nice to link the commit which was being deployed

DEBIAN/control
```diff
@@ -0,0 +1,8 @@
+Package: pyrsia
```
prince-chrismc, line 1: For pretty 💄 can this be in a subfolder?

.github/workflows/rust.yml
```diff
@@ -29,6 +29,84 @@ jobs:
...
+    - uses: actions/upload-artifact@v2
+      if: ${{ github.repository_owner == 'pyrsia' && runner.os == 'Linux' }}    
+      with:
+        name: DEBIAN
+        path: DEBIAN
+
+# Create the .deb file, add it to the apt repo and
+# push the changes to the website github repo
+  apt-package:
+      if: ${{ github.repository_owner == 'pyrsia' }}
```
prince-chrismc, line 56: We should block this on PRs too... they  wont have the secrets

prince-chrismc (commented):
Light review with some general questions/feedback

.github/workflows/rust.yml
sbtaylor15: The DEBIAN directory is used by the debpkg program to generate the .Deb file.  I put it in the main repo instead of the website repo to keep it in sync the the code.  More files will be added when we get the service added.

.github/workflows/rust.yml
sbtaylor15: We would need to modify the action to add different message.  We should make the action change a different issue.

DEBIAN/control
sbtaylor15: This is the part of the name of the .deb file.  Also used when doing the install. `apt-get install pyrsia`.  Not related to directory names.

.github/workflows/rust.yml
sbtaylor15: What is the variable to check if it's a PR?

prince-chrismc:
```suggestion
      if: github.event_name == 'push' && github.repository_owner == 'pyrsia' #PR's nor forks need to upload 
```

DEBIAN/control
prince-chrismc: I understand what it's for but it would be great (to me) to have the root of the project more "presentable".

Perhaps a small rename like `installer/DEBIAN/control` so we can group all the files (I expect we'll need more not less)

.github/workflows/rust.yml
prince-chrismc: Yes, thats fine.

My curiosity is on the upload and download that's being done (instead of a checkout in the next job) for this particular artifact as you point out 

> in the main repo

So the build does not affect it (that I know of)

.github/workflows/rust.yml
prince-chrismc: Created https://github.com/sbtaylor15/apt-repo-action/pull/1 to track the "issue"

prince-chrismc:
I also discovered https://crates.io/crates/cargo-deb 

> This is a Cargo helper command which automatically creates binary Debian packages (.deb) from Cargo projects.

Did you see this when you looked into this?

.github/workflows/rust.yml
sbtaylor15: The checkouts wipeout the working directory so you need to persist the binaries between jobs by uploading them somewhere and then downloading in the other job.  This is similar to the actions/cache@v2 but has more control.

DEBIAN/control
sbtaylor15: Ah, I understand now.  Yes, we can create an installer folder.

sbtaylor15:
I did not see the cargo-deb.  Let me take a look at that to see if it simplifies some of the steps.

prince-chrismc:
> I did not see the cargo-deb. Let me take a look at that to see if it simplifies some of the steps.

Let's make it a new issue? a follow up to this? I'd love to have this in main so we can workout the kinks

sbtaylor15:
> https://crates.io/crates/cargo-deb

See issue #129

prince-chrismc:
One last comment (and the PR CI will pass) https://github.com/pyrsia/pyrsia/pull/116#discussion_r770913928

prince-chrismc (approved):

prince-chrismc (dismissed):

prince-chrismc:
Does this also close #59 ?

sbtaylor15:
> 

Yes

.github/workflows/rust.yml
```diff
@@ -29,6 +29,84 @@ jobs:
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: DEBIAN
+          path: pyrsia/DEBIAN
+
+# create the pyrsia.deb file
+      - name:  Package for apt-get
+        run: |
+            mkdir -p pyrsia/usr/local/bin; \
+            cp target/debug/pyrsia_node pyrsia/usr/local/bin; \
```
chb0github, line 83: Is there a way you can insert an error exit directive in here? `set -e` ? If any of these fail, the others still run and you will get the error code of the last command: `ls -la prysia.deb`

chb0github (approved):

.github/workflows/rust.yml
prince-chrismc: With #129 this should go away 🤞  but good suggestion

.github/workflows/rust.yml
```diff
@@ -98,7 +98,7 @@ jobs:
             bionic
           repo_supported_arch: |
             amd64
-          page_branch: main
+          page_branch: gh-pages
```
prince-chrismc, line 101: I'd of preferred main but no objection

prince-chrismc (approved):

prince-chrismc:
https://github.com/pyrsia/pyrsia/runs/4609127711?check_suite_focus=true#step:3:26
Failed check is fixed in #140

PR #128, opened 2021-12-17
First cut of a blockchain

### LINKED ISSUE #56: Create a mock blockchain
Create a mock blockchain manager suitable for testing internal clients of the blockchain managers.

chb0github:
The error I get:

```bash
error[E0515]: cannot return reference to temporary value
  --> pyrsia_node/src/block_chain/block_chain.rs:34:9
   |
34 |            &mut BlockChain {
   |  __________^____-
   | | _________|
   | ||
35 | ||             blocks: vec![]
36 | ||         }
   | ||         ^
   | ||_________|
   | |__________returns a reference to data owned by the current function
   |            temporary value created here

warning: variable does not need to be mutable
   --> pyrsia_node/src/main.rs:215:9
    |
215 |     let mut bc = BlockChain::new();
    |         ----^^
    |         |
    |         help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

```
resolves #56


chb0github:
I'd like to get a compiling integration checked in, and then immediately iterate on doing it better. 

I full accept my understanding of blockchain is incomplete

prince-chrismc:
the chrono library has a vulnerability, https://github.com/pyrsia/pyrsia/runs/4564598702?check_suite_focus=true#step:3:26 try using the time crate directly

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct BlockChain {
+    blocks: Vec<Block>,
+}
+
+
+const DIFFICULTY_PREFIX: &str = "00";
+
+impl BlockChain {
+    pub fn new() -> &'static mut Self {
+        &mut BlockChain {
```
prince-chrismc, line 34: Typically the `new` should look like this

```suggestion
    pub fn new() -> Self {
        BlockChain {
```

When you instantiate something you _always_ expect the caller to take ownership.

The new `BlockChain` instance allocated on line 34 will be destroyed on line 37 so the caller can not have a reference to the memory (that the compiler is trying to return).

pyrsia_node/src/main.rs
```diff
@@ -61,6 +62,8 @@ use std::{
 use tokio::io::{self, AsyncBufReadExt};
 use tokio::sync::Mutex;
 use warp::Filter;
+use crate::block_chain::block::Block;
+use crate::block_chain::block_chain::BlockChain;
```
prince-chrismc, line 66: I think the `crate` is extra since we have the `mod` 🤔 

```suggestion
use block_chain::{block::Block, block_chain::BlockChain};
```

prince-chrismc:
1. Please give your PR a more descriptive title... something like "Outline for a mock blockchain"
2. Can you link the issue (so that GitHub knows what to do). You can check out https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

pyrsia_node/src/main.rs
chb0github: It wouldn't compile without, and my editor suggested it. I'll try again

pyrsia_node/src/block_chain/block_chain.rs
chb0github: I orginally had this but that gave me other issues. Will retry

chb0github:
@prince-chrismc - by putting GH-XXX in the title, GH seems to link to it - but you are saying an explicit URL in the title?

pyrsia_node/src/block_chain/block_chain.rs
mgrand: @chb0github The problem is that you are returning a reference to the struct you instantiated. The struct itself has to go away when it goes out of scope.  Return the struct, not a reference to it, and then ownership moves to the caller.

prince-chrismc:
I mean linking it so that Github can automatically close the issue when the PR is merged,
The textual link is automatically added in the commit messages when merging (you dont need to add it manually here)

prince-chrismc:
Please consider https://github.com/chb0github/pyrsia/pull/1

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,153 @@
...
+            data: String::from("genesis!"),
+            nonce: 2836,
+            hash: "0000f816a87f806bb0073dcf026a64fb40c946b5abee2573702828694d5b4c43".to_string(),
+        };
+        self.blocks.push(genesis_block);
+        self
+    }
+    pub(crate) fn dump(&self) -> Result<String, Error> {
+        serde_json::to_string_pretty(&self.blocks).map(|pretty_json| {
+            println!("{}", pretty_json);
```
prince-chrismc, line 52: We have been trying to use the logging function over print to console 😄

pyrsia_node/src/block_chain/block_chain.rs
chb0github: Yeah, I spend hours this weekend aligning my mind to ownership. I fixed it. It's not as I want it to be, but it works.

pyrsia_node/src/main.rs
chb0github:
```
215 |     let mut bc = BlockChain::new();
    |                  ^^^^^^^^^^ not found in this scope
    |
help: consider importing this struct
    |
37  | use crate::block_chain::block_chain::BlockChain;
    |

error[E0422]: cannot find struct, variant or union type `Block` in this scope
   --> pyrsia_node/src/main.rs:259:37
    |
259 |                         let block = Block{
    |                                     ^^^^^ not found in this scope
    |

```
When I do as you suggest

pyrsia_node/src/block_chain/block_chain.rs
chb0github: Done.

pyrsia_node/src/main.rs
prince-chrismc: Your right.

I opened a PR but did not change my suggestion here... correct code is https://github.com/chb0github/pyrsia/pull/1/files#diff-2c5d9cbb9033cbd429ad804a903684afdac3c2f5cb96bd0c474b56f0793521fbR39-R40

Thanks for running CI, it passed https://github.com/chb0github/pyrsia/actions/runs/1594267461 🚀

pyrsia_node/src/main.rs
```diff
@@ -66,6 +66,7 @@ use std::{
 use tokio::io::{self, AsyncBufReadExt};
 use tokio::sync::Mutex;
 use warp::Filter;
+use super::block_chain::{block::Block, block_chain::BlockChain};
```
prince-chrismc, line 69: I think the reason it's failing is this new line you added. 

Bad communication. I should have resolved my suggestion but everything was good with my PR 

try removing it

```suggestion
```

pyrsia_node/Cargo.toml
```diff
@@ -44,6 +44,7 @@ tokio = { version = "1", features = [ "macros", "rt-multi-thread", "io-std" ] }
 unqlite = "1.5.0"
 uuid = { version = "0.8.2", features = [ "v4" ] }
 warp = { version = "0.3.1", default-features = false }
+chrono = "0.4"
```
prince-chrismc, line 47: dont forget to remove this =)

mgrand:
@chb0github The first payload that we will need the blockchain to have is identities. There will be others. The data in the block should be an array of JSON objects.

The identity struct is currently defined in pyrsia_node::node_manager::model::identity. (model will be moving to pyrsia_client_lib)

Because it is annotated with 
`#[signed_struct]`
it implements the `Signed` trait. This trait, with the help of the  `#[signed_struct]` macro, provides some methods you will be interested in:
* `json` returns the signed json associated with the struct.
* `from_json_string` is a static method to create a struct from signed JSON
* `sign_json` creates the JSON representation of the struct and adds a signature. If there is already signed JSON, then it just adds a signature.
* `verify_signatures` verifies the signatures in a struct's JSON and returns information about the signatures.

Identities provide information about public signing key that identifies the signing key.

I assume that you will eventually annotate the Block struct with `#[signed_struct]` so that commit nodes can sign the blocks. This will allow 

I have identified at least two signing keys that will need to be put in the genesis block as part of blockchain creation:
* Blockchain blocks should be signed so that when a non-consensus non-commit node receives a new blockchain block it can verify the the block has a valid signature. The identity used by the commit nodes to sign the blocks should be in the genesis block with some metadata that indicates it is authorized to sign committed blocks.
* The metadata manager is going to have some built-in metadata. This will need to be signed by us. The identity we use to sign it should also be in the genesis block.

fishseabowl:
I agree with Mark's comment. At the beginning, we could only support centralized commit nodes.

prince-chrismc (approved):
LGTM!

Please sign the CLA and update your title 👍

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+            timestamp: now(),
+            previous_hash: String::from("genesis"),
+            data: String::from("genesis!"),
+            nonce: 2836,
+            hash: "0000f816a87f806bb0073dcf026a64fb40c946b5abee2573702828694d5b4c43".to_string(),
+        };
+        self.blocks.push(genesis_block);
+        self
+    }
+    pub(crate) fn dump(&self) -> Result<String, Error> {
```
prince-chrismc, line 46: Is the crate context needed?

```suggestion
    pub fn dump(&self) -> Result<String, Error> {
```

pyrsia_node/src/main.rs
```diff
@@ -250,6 +256,19 @@ async fn main() {
                     cmd if cmd.starts_with("get_blobs") => {
                         swarm.behaviour_mut().lookup_blob(message).await
                     }
+                    "block" => {
+                        // assuming the message is a json version of the block
+
+                        let block = Block {
```
MitaliBo, line 262: i am assuming this is just a template because we are not doing anything with message?

MitaliBo (approved):

prince-chrismc (changes requested):
Blocking because the CLA is not being required

prince-chrismc (approved):
Nevermind, just some bad image caching

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+    blocks: Vec<Block>,
+}
+
+const DIFFICULTY_PREFIX: &str = "00";
+
+impl BlockChain {
+    pub fn new() -> Self {
+        BlockChain { blocks: vec![] }
+    }
+    pub fn genesis(&mut self) -> &mut Self {
```
mgrand, line 34: Should the genesis function also have crate scope? It sounds like something that would be called internally as part of blockchain initialization. It is not clear that there is a use for external calls to it.

pyrsia_node/src/block_chain/block_chain.rs
mgrand: @prince-chrismc Why are you concerned about the crate scope? This appears to be a method to use for dumping out the contents of the blockchain. At this point, it is perfectly reasonable to not want to expose it as part of the Blockchain's public API.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+            pretty_json
+        })
+    }
+    pub fn add_block(&mut self, block: Block) -> Option<&Block> {
+        // let block = self.mk_block( data.clone()).expect("error creating block");
+        let last_block = self.blocks.last().expect("has a block");
+        if self.is_block_valid(&block, last_block) {
+            self.blocks.push(block);
+            return self.blocks.last();
+        }
```
mgrand, line 58: @chb0github This raises the interesting question of what should our strategy be if a node has an invalid blockchain. Would you like to organize a team discussion about that, or would you rather I do it?

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+        self
+    }
+    pub(crate) fn dump(&self) -> Result<String, Error> {
+        serde_json::to_string_pretty(&self.blocks).map(|pretty_json| {
+            info!("{}", pretty_json);
+            pretty_json
+        })
+    }
+    pub fn add_block(&mut self, block: Block) -> Option<&Block> {
+        // let block = self.mk_block( data.clone()).expect("error creating block");
```
mgrand, line 53: Did you mean to leave this commented line in the code?

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+            return self.blocks.last();
+        }
+        None
+    }
+    pub fn mk_block(&mut self, data: String) -> Option<Block> {
+        let now = now();
+        self.blocks
+            .last()
+            .map(|last_block| {
+                let (nonce, hash) = mine_block(last_block.id, now, &last_block.hash, &data);
```
mgrand, line 66: We do not need to mine blocks. If there is consensus on all of the transactions in the data (which should be an array of JSON objects) then that should be sufficient. The signing format we are using allows for multiple signatures to be present for a single JSON object. Proof of consensus could be the presence of a signature (or signatures?) on a JSON object whose public key (the public key is part of the signature) identifies an identity on the blockchain that includes an attribute identifying it as authorized to certify consensus.

pyrsia_node/src/block_chain/block.rs
```diff
@@ -0,0 +1,26 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Block {
+    pub id: u64,
+    pub hash: String,
+    pub previous_hash: String,
+    pub timestamp: u128,
+    pub data: String,
+    pub nonce: u64,
```
mgrand, line 25: Since we are not mining (everything we are recording on the blockchain has an independent existence outside of the blockchain), it is not clear that there is a need for a nonce.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+                previous_hash: last_block.hash.clone(),
+                timestamp: now,
+                data,
+                nonce,
+            })
+    }
+
+    fn is_block_valid(&self, block: &Block, previous_block: &Block) -> bool {
+        if block.previous_hash != previous_block.hash {
+            warn!("block with id: {} has wrong previous hash", block.id);
```
mgrand, line 81: If code detects a situation for which I do not have a way to recover or continue normal operations, I usually call that an error rather than a warning.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+            return false;
+        }
+        if !hash_to_binary_representation(&hex::decode(&block.hash).expect("can decode from hex"))
+            .starts_with(DIFFICULTY_PREFIX)
+        {
+            warn!("block with id: {} has invalid difficulty", block.id);
+            return false;
+        }
+        if block.id != previous_block.id + 1 {
```
mgrand, line 90: This is OK for now. The the production blockchain, I would like to see the sequence of blocks be based on physical ordering in storage rather than being part of the blockchain's internal data structure. Keeping the block sequence dependent on something external to the blockchain structure is more secure.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+            block.nonce,
+        )) != block.hash
+        {
+            warn!("block with id: {} has invalid hash", block.id);
+            return false;
+        }
+        true
+    }
+
+    fn is_chain_valid(&self, chain: &[Block]) -> bool {
```
mgrand, line 111: I don't see a call to this function. What is the policy for when this should be called and what to do if it returns false?

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        true
+    }
+}
+
+fn now() -> u128 {
+    match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(n) => n.as_millis(),
+        Err(_) => panic!("SystemTime before UNIX EPOCH!"),
```
mgrand, line 126: I don't see why system time before the UNIX epoch is a reasonable threshold for panic. If the concern is that the time predates Pyrsia, then the panic threshold should be if it is before this year.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+fn mine_block(id: u64, timestamp: u128, previous_hash: &str, data: &str) -> (u64, String) {
```
mgrand, line 130: We will rely on signatures rather than mining.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -0,0 +1,174 @@
...
+    nonce: u64,
+) -> Vec<u8> {
+    let data = serde_json::json!({
+        "id": id,
+        "previous_hash": previous_hash,
+        "data": data,
+        "timestamp": timestamp,
+        "nonce": nonce
+    });
+    let mut hasher = Sha256::new();
```
mgrand, line 166: I would like to see us use a more secure hash. Please consider BLAKE3. It requires less effort to compute than SHA3-512 and is about as secure.

pyrsia_node/src/main.rs
```diff
@@ -210,7 +215,8 @@ async fn main() {
 
     tokio::spawn(server);
     let tx2 = tx.clone();
-
+    let mut bc = BlockChain::new();
+    bc.genesis();
```
mgrand, line 219: Shouldn't the decision to create a genesis block be internal to the blockchain? Perhaps new should call it if there is no local blockchain.

pyrsia_node/src/main.rs
mgrand: I assume this is a temporary mechanism for pushing blocks to a node.

mgrand (approved):
I had a few comments.

pyrsia_node/src/main.rs
chb0github: Yeah, this is just a "hey, insert block from another node, here"

pyrsia_node/src/block_chain/block_chain.rs
chb0github: I actually wanted it to be part of the `new`, but my limited rust was getting in the way. I definitely think it should be cooked in

pyrsia_node/src/block_chain/block_chain.rs
chb0github: Rallying the troops at this point will be on me. There are any number of strategies for reconcilliation - some short term and focused on MVP and others more production grade.

pyrsia_node/src/block_chain/block_chain.rs
chb0github: no. But please don't make me change it - then I would need 3 approvals again

pyrsia_node/src/block_chain/block_chain.rs
chb0github: Once we have a more concrete block we all agree on, I'll happily change it. This was just for demonstration purposes.

pyrsia_node/src/block_chain/block_chain.rs
chb0github: Ok. Then we'll have to talk about the implementation.

pyrsia_node/src/block_chain/block_chain.rs
chb0github: This was ripped right out of rust example docs - the reason they return an `either/Result` is because you can't count on the system time to be monotonically increasing. I mean, I can change this to `.expect("we went back in time")`

pyrsia_node/src/block_chain/block_chain.rs
chb0github: Obviously we need a consensus phase. I am taking down a lot of notes for this next iteration to resolve

pyrsia_node/src/main.rs
chb0github: Yes. Again, my limited rust was getting in the way. When you call `BlockChain::new` I expect the genesis block is in place.

chb0github:
I have taken notes and will address each of these inputs - Thanks all.

PR #132, opened 2021-12-20
Only run label action on pyrsia and not on forks

### LINKED ISSUE #130: auto-label action failing on forked repo
Should we add an if: to only run on repos owned by pyrsia (`github.repository_owner == 'pyrsia'`) ?

image

sbtaylor15:
Resolves #130

### DIFF
--- a/.github/workflows/label-conflicts.yml
+++ b/.github/workflows/label-conflicts.yml
@@ -10,6 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: prince-chrismc/label-merge-conflicts-action@v1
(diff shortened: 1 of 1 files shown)


prince-chrismc:
I did not know this perviously. So I'll share here incase anyone else wonders how this comes about.

If you fork a repository you get all of it's labels.
If upstream adds new labels.
Forks are unaffected.

Because forks were made before the new label was add and used in Actions, they don't have them so the action fails with "missing label".

prince-chrismc (approved):

mgrand (approved):

PR #136, opened 2021-12-20
Metadata manager implementation

mgrand:


pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
         todo!()
     }
 }
+
+// Names of document stores and their indexes
+const DS_PACKAGE_TYPES: &str = "package_types";
+const IX_PACKAGE_TYPES_NAMES: &str = "names";
```
prince-chrismc, line 192: Abbreviated things are very hard to understand. I assume `IX` is index but that seems like a stretch (my brain saw roman numerals)

Two suggestions for this one

1. Just use the full name
2. Make it an `enum` Some like this https://stackoverflow.com/a/65040451

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
...
+    }
+}
+
+/// Open the specified document store. If that fails, try creating it.
+/// * `ds_name` ― The name of the document store.
+/// * `index_specs` ― If the document store need to be created, this method is called to get a
+///                   a description of the indexes it will have.
+/// * `initial_records` ― If the document store needs to be created, this method is called to get
+///                       JSON string that will each be inserted as a record into the document store.
+fn open_document_store(
```
prince-chrismc, line 274: Shouldn't this logic belong to the document store module? I imagine everything that uses it will need to duplicate this code you added 🤔

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
...
+            Ok(_) => insert_metadata(&self.namespace_docs, namespace),
+            Err(error) => untrusted_metadata_error(namespace, &error.to_string()),
+        }
+    }
+
+    fn get_namespace(
+        &self,
+        package_type: PackageTypeName,
+        namespace_path: &str,
+    ) -> anyhow::Result<Option<Namespace>, anyhow::Error> {
```
prince-chrismc, line 362: According to the `anyhow` docs this is the same thing.

```suggestion
    ) -> anyhow::Result<Option<Namespace>> {
```

Is there a reason you choose to add the default error type in?

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
...
+mod tests {
+    use super::*;
+    use pyrsia_client_lib::signed::{create_key_pair, JwsSignatureAlgorithms};
+    use rand::Rng;
+    use serial_test::serial;
+    use std::path::Path;
+
+    const DIR_PREFIX: &str = "metadata_test_";
+
+    fn do_in_temp_directory(
```
prince-chrismc, line 547: Might be a good spot for the tempfile crate we discussed this morning?

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
...
+        Err(error) => {
+            warn!(
+                "Failed to open document store {}; error was {}",
+                ds_name, error
+            );
+            info!("Attempting to create document store {}", ds_name);
+            match DocumentStore::create(ds_name, index_specs()) {
+                Ok(ds) => {
+                    info!("Created document store {}", ds_name);
+                    populate_with_initial_records(&ds, initial_records)?;
```
prince-chrismc, line 291: Whats the motivation for this?

the "initial" records are always empty, it's very out of place 😕

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
...
+                "Failed to insert initial record into document store {}: {}\nError was {}",
+                ds.name, record, error
+            );
+            todo!("If an attempt to insert an initial record into document store fails, then we need to do something so that we will know that the document store is missing necessary information")
+        }
+    }
+    Ok(())
+}
+
+fn failed_to_create_document_store(
```
prince-chrismc, line 316: Function called only once? Are you expecting to need this in lots of places?

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +186,462 @@ impl Iterator for PackageIterator {
...
+        match self.trust_manager.trust_package_type(pkg_type) {
+            Ok(_) => insert_metadata(&self.package_type_docs, pkg_type),
+            Err(error) => untrusted_metadata_error(pkg_type, &error.to_string()),
+        }
+    }
+
+    fn get_package_type(
+        &self,
+        name: PackageTypeName,
+    ) -> anyhow::Result<Option<PackageType>, anyhow::Error> {
```
prince-chrismc, line 339: Why would we be looking up a non-existant package type?

My expectation is this is set at the application level (we only need 1 or 2). So why is this being persisted on the users disk in the first place?

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: Thank you for pointing out that these names are hard to understand. There is a naming convention I am using that clearly requires more explanation then my one line comment. I will expand that.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: The document store is all about how to store, retrieve and soon update JSON documents. It has no idea about what to store. The first time the metadata manager uses the document store, it has to tell the document store what collections it needs to create and how they are to be indexed.

The document store is not intended to be specific to the metadata manager. With a small extension to the way it organizes indexes, it could be used by the blockchain to store blocks, indexing the JSON objects in each block so that we can quickly find an identity that is stored on the blockchain.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: No. I should simplify the declarations as there is no advantage to the more complex form for the default case. Thank you for pointing this out.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: This does not create temp files. It handles the problem of wanting to clean up files created by the document store during unit tests by creating a directory with a random name, changing the current directory to the new directory, running a test, setting the current directory back to its previous value and then deleting the directory after then test has run. 

The reason I chose this approach is that I did not want any different logic or names used for creating document store files during unit tests. Since UnQLite just uses the current directory for its files this works well for this purpose.

It does have a drawback. Because there is only one current directory, the tests cannot run in parallel. I have to force them to run serially.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: The initial records are empty because we have not created them yet. That will be for a future issue/task. For each type of package that we support, we will need a PackageType record. Right now we just need one record for Docker packages. Part of the process of supporting a new type of package will be to add a record that defines the corresponding package type.

Since this record will be a JSON object that needs to be signed, creating it will have to wait for us to create a stand-alone signing tool for predefined JSON. We will also need such a tool to put predefined identities in the blockchain. For example, the identity that signs the PackageType JSON will need to be pre-installed in the blockchain so that it can be trusted.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: I sometimes look at a larger piece of code and refactor it by making a block a separate function. Doing this gives the block a descriptive name. the possibly one place that calls it becomes easier to understand because its structure is simpler and named.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: I don't expect it to happen. I write code so that it has a reasonable way to deal with the obvious ways it can be used and abused. I am checking this for all types of metadata, not just package types.

The two  most likely ways this can happen is 
* There is a problem with the document store or the files it relies on. In this case it is important to have good log entries to help diagnose the problem.
* In the future, when we are adding support for a new package type, a node that has not yet been upgraded may be asked for information about a package type is does not have defined. 
* One other way this could happen is a hacker just looking for weaknesses.

pyrsia_node/src/metadata_manager/metadata.rs
mgrand: I have added a comment to `populate_with_initial_records` with this information.

mgrand:
closing this because I am now working from a fork rather than this branch.

PR #139, opened 2021-12-21
Adds cargo-make and a ci flow to be able to run the same ci on each machine

betarelease:


betarelease:
Since cargo-make has a dependency `chrono` that is failing audit checks I am closing this PR at this time. Will create an issue for this and send a fresh PR when there is an updated version of cargo-make.

Issue #71, opened 2021-12-06
"pyrsia status" shows node count, artifact count and disk available


MitaliBo:

prince-chrismc:
@MitaliBo You moved this to "done" but did not close the issue

Was this completed with #140 ?

MitaliBo:
Yes handled #140

PR #140, opened 2021-12-21
cli command: `pyrsia status` implementation

### LINKED ISSUE #112: pyrsia_cli to call pyrsia_node_api for getting data for "node ls" and "status"

MitaliBo:
Implementation of status command to show : total_artifacts, peers_connected_count and disk_space_available

closes #112 #71


pyrsia_cli/src/main.rs
```diff
@@ -34,23 +36,22 @@ async fn main() {
             if config_matches.is_present("add") {
                 let node_config = config_matches.value_of("add").unwrap();
                 let _result = add_config(String::from(node_config));
-                println!("Node configured: {}", node_config);
+                println!("Node configured:      {}", node_config);
```
mgrand, line 39: Please use the log API. These look link info!() or debug!() messages.

mgrand (approved):

pyrsia_cli/src/main.rs
MitaliBo: No these are the output of cli commands and logs does not output them in command line

prince-chrismc:
Security audit failed https://github.com/pyrsia/pyrsia/runs/4598120109?check_suite_focus=true#step:3:30

MitaliBo:
> https://github.com/pyrsia/pyrsia/runs/4598120109?check_suite_focus=true#step:3:30

yeah i am quite confused i had security audit passed before when i resolved conflict changes and merged,  it failed, tracking where this is coming from, but it should have failed in the previous PR which merged

pyrsia_cli/src/commands/node.rs
```diff
@@ -28,7 +29,7 @@ pub async fn ping() -> Result<String, reqwest::Error> {
             println!("Error: {}", error);
         }
     };
-
+    //TODO: implement ping api in kademlia
```
prince-chrismc, line 32: I thought Kademlia is for nodes talking to each other. 🤔 

Is the comment still relevant?

pyrsia_node/Cargo.toml
```diff
@@ -23,8 +23,8 @@ hyper = { version = "0.14", features = ["full"] }
 hyper-tls = "0.5.0"
 itertools = "0.10.3"
 lazy_static = "1.4.0"
-#libp2p = "0.40.0"
-libp2p = {git = "https://github.com/decentnetwork/rust-libp2p.git", branch = "fix-compiling-v0.40", features=["tcp-tokio"]}
+#libp2p = "0.42.0"
+libp2p = {git = "https://github.com/decentnetwork/rust-libp2p.git", features=["tcp-tokio"]}
```
prince-chrismc, line 27: We should remove the `git` to the fork. Look like it's just synced to the upstream.

Theres no benefit

pyrsia_node/src/artifact_manager.rs
```diff
@@ -275,6 +282,29 @@ impl<'a> ArtifactManager {
         Err(anyhow!("Not an accessible directory: {}", repository_path))
     }
 
+    pub fn artifacts_count(&self, repository_path: &str) -> Result<usize, Error> {
+        let mut total_files = 0;
+
+        for file in WalkDir::new(repository_path)
+            .into_iter()
+            .filter_entry(|path| Self::is_not_hidden(path))
```
prince-chrismc, line 290: Why exclude hidden files?

pyrsia_node/src/node_api/handlers/swarm.rs
```diff
@@ -15,6 +15,10 @@
 */
 
 // this is to handle calls from cli that needs access info swarm specific from  kad dht
+use super::super::Status;
```
prince-chrismc, line 18: Extra given the line below?

pyrsia_node/src/node_api/handlers/swarm.rs
```diff
@@ -39,3 +43,40 @@ pub async fn handle_get_peers(
...
+pub async fn handle_get_status(
+    tx: Sender<String>,
+    rx: Arc<Mutex<Receiver<String>>>,
+) -> Result<impl Reply, Rejection> {
+    match tx.send(String::from("peers")).await {
+        Ok(_) => debug!("request for peers sent"),
+        Err(_) => error!("failed to send stdin input"),
+    }
+
+    let peers = rx.lock().await.recv().await.unwrap();
```
prince-chrismc, line 56: Amazing work finding this 👏

prince-chrismc (commented):
Some questions and minor suggestions

pyrsia_cli/src/main.rs
```diff
@@ -14,43 +14,113 @@
...
-use clap::{load_yaml, App};
+use clap::{crate_authors, crate_description, crate_version, App, AppSettings, Arg};
 
 #[tokio::main]
 async fn main() {
-    let yaml = load_yaml!("cli.yaml");
-    let matches = App::from(yaml).get_matches();
+    //parsing command line arguments
+
+    let matches = App::new("pyrsia")
```
chb0github, line 32: Yeah, I been thinking about this code: Can this please be broken down into functions or something? It's way too verbose. Or better yet, can we do some kind of discovery with `[command]` such things?

pyrsia_cli/src/main.rs
```diff
@@ -14,43 +14,113 @@
...
-                        println!("Error: {}", error);
+                        println!("No Node Configured:       {}", error);
                     }
                 };
             }
         }
 
-        //node subcommand
         Some(("node", node_matches)) => {
             if node_matches.is_present("ping") {
```
chb0github, line 125: I believe a `match` expression, instead of all the `if/else` would make this tidier

pyrsia_node/src/artifact_manager.rs
chb0github: your last `if` is really just another filter: `filter(|file| file.metadata().unwrap().is_file()).count()`

This mix of loop and functional is a bit odd

pyrsia_node/src/artifact_manager.rs
```diff
@@ -275,6 +282,29 @@ impl<'a> ArtifactManager {
...
+            }
+        }
+        Ok(total_files)
+    }
+
+    fn is_not_hidden(entry: &DirEntry) -> bool {
+        entry
+            .file_name()
+            .to_str()
+            .map(|s| entry.depth() == 0 || !s.starts_with('.'))
```
chb0github, line 304: This is by convention - on windows a file is `hidden` if windows says so. I assume there to be an api you can invoke to get this information

chb0github: wait -- a directory of size `0` is hidden?

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -170,5 +170,5 @@ fn calculate_hash(
 fn hash_to_binary_representation(hash: &[u8]) -> String {
     hash.iter()
         .map(|c| format!("{:b}", c))
-        .fold("".to_string(), |cur, nxt| cur + &nxt)
+        .fold("".to_string(), |cur, nxt| format!("{}{}", cur, nxt))
```
chb0github, line 173: Why?

pyrsia_node/src/main.rs
```diff
@@ -252,13 +259,15 @@ async fn main() {
...
                     "peers" => swarm.behaviour_mut().list_peers_cmd().await,
-                    _ => match tx2.send(line).await {
+                    _ => match tx4.send(line).await {
                         Ok(_) => debug!("line sent"),
                         Err(_) => error!("failed to send stdin input"),
                     },
                 },
                 EventType::Message(message) => match message.as_str() {
-                    "peers" => swarm.behaviour_mut().list_peers(local_peer_id).await,
+                    cmd if cmd.starts_with("peers") || cmd.starts_with("status") => {
```
chb0github, line 268: Instead of the `if` statements in the match, can you just flatten out the match? This main loop is getting really long

chb0github (commented):
I have some suggestions, some headscratchers and some things I think should be changed. The functions here in are getting really long and verbose. For the commands, some kind of discovery mechanism would be great

pyrsia_node/Cargo.toml
mgrand: @prince-chrismc There is a benefit. 42 is not yet released.

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -170,5 +170,5 @@ fn calculate_hash(
 fn hash_to_binary_representation(hash: &[u8]) -> String {
```
mgrand, line 170:
```suggestion
fn hash_to_binary_string_representation(hash: &[u8]) -> String {
```

pyrsia_node/src/block_chain/block_chain.rs
MitaliBo: after upgrading to new version of libp2p some of the dependencies pulled in which deprecated the use of '+', anyhow in rust i thought they never supported concatenate with '+' not sure how you were able to use that

pyrsia_node/src/main.rs
```diff
@@ -127,7 +125,7 @@ async fn main() {
     // Reach out to another node if specified
     if let Some(to_dial) = matches.value_of("peer") {
         let addr: Multiaddr = to_dial.parse().unwrap();
-        swarm.dial_addr(addr).unwrap();
+        swarm.dial(addr).unwrap();
```
mgrand, line 128: This will panic if dial fails. We should explicitly check for the failure.

pyrsia_cli/src/main.rs
MitaliBo: it is verbose and it cannot break down because it defines the cli structure, previously i had implemented yaml method which was better but last week they deprecated that and i have to fallback to builder pattern

mgrand (dismissed):
Left a few comments.

pyrsia_node/src/artifact_manager.rs
MitaliBo: i am thinking we won't have artifacts with hidden files and for christian answer one filter is filtering the hidden files and other is checking if its a file not directory in that tree

pyrsia_node/src/node_api/handlers/swarm.rs
MitaliBo: extra what?

pyrsia_node/Cargo.toml
MitaliBo: nope libp2p standard version still does not work so still have to use fork

pyrsia_cli/src/commands/node.rs
MitaliBo: sorry its particularly for node daemon not kademlia

pyrsia_node/src/artifact_manager.rs
chb0github: My comment remains: You're mixing styles

pyrsia_cli/src/main.rs
chb0github: You mean, you can define a function called `FooCommand::args()` that just returns the necessary args to feed this, and isolate those details into an entirely different file?

Forgive me, but a function that stretches hundreds of lines (with no end in sight) is not something I am comfortable with. We need to find a more maintainable way to add args

pyrsia_node/src/block_chain/block_chain.rs
chb0github: You know, I also noticed functions like `.join()` show up in intellisense but fail to compile. I am very much confused on what is and is not supported

pyrsia_node/src/block_chain/block_chain.rs
chb0github: The naming change seems redundant - it returns a string

pyrsia_node/src/main.rs
chb0github: And do what? Panic and exit?

pyrsia_node/src/artifact_manager.rs
MitaliBo: i would be happy to take suggestion on how to embed it in loop, i had tried but that does not work

pyrsia_node/src/main.rs
MitaliBo: @mgrand we want to panic if we are not able to dial the peer, it makes no sense to proceed :)

pyrsia_node/src/block_chain/block_chain.rs
MitaliBo: may be you need to update your rust-analyzer or whatever plugin you are using in intellij

pyrsia_cli/src/main.rs
MitaliBo: oh if you meant the whole cli structure, yes that i am moving, i thought you meant breaking down the structure into different functions

pyrsia_node/src/main.rs
mgrand: What does the CLI user do? Shouldn't they see a message saying that we were unable to dial the peer rather than that we crashed?

mgrand (approved):

pyrsia_cli/src/commands/node.rs
prince-chrismc:
```suggestion
    //TODO: implement ping api in Node
```

pyrsia_node/Cargo.toml
prince-chrismc: 🤢

pyrsia_node/src/node_api/handlers/swarm.rs
prince-chrismc: extra import sorry, missed a word.

```suggestion
```

pyrsia_node/src/artifact_manager.rs
MitaliBo: i am sure if i understand it correctly but its not the size but the depth of directory

prince-chrismc (approved):

pyrsia_node/src/main.rs
prince-chrismc: This is the node (not the CLI) and it's just example code that's out of scope of Demo.

pyrsia_node/Cargo.toml
prince-chrismc: Sorry for picking on this issue.

image

This fork is just slightly behind the "real" one so this seems a little over kill. 

Ill open a new PR since it's completely unrelated.

***

Amazing work figuring out the problem once again

prince-chrismc: https://github.com/pyrsia/pyrsia/issues/151

mgrand (approved):

PR #143, opened 2021-12-22
Metadata implementation

mgrand:
This is an initial implementation of the metadata manager. All features for the demo are implemented, except for those that require the document store to support range searches or updates.


pyrsia_node/src/node_manager/model/artifact.rs
```diff
@@ -44,3 +44,59 @@ pub struct Artifact {
...
+            creation_time,
+            url,
+            size,
+            mime_type,
+            metadata,
+            source_url,
+        }
+    }
+
+    pub fn hash(&self) -> &Vec<u8> {
```
chb0github, line 73: I don't understand this code - struct fields can't be made private (I don't think) - so why have all these public getters?

chb0github (commented):
There is 1 bit of code I don't get.

pyrsia_node/src/node_manager/model/artifact.rs
mgrand: @chb0github I want instances of this struct to be immutable. Struct fields are private by default. They can be made public.

pyrsia_node/src/node_manager/model/package_type.rs
```diff
@@ -19,16 +19,18 @@ extern crate serde_json;
 
 use pyrsia_client_lib::signed::Signed;
 use serde::{Deserialize, Serialize};
-
 use signed_struct::signed_struct;
+use std::string::ToString;
```
prince-chrismc, line 23: I dont see `ToString` being used 🤔 

```suggestion
```

pyrsia_node/src/document_store/document_store.rs
```diff
@@ -372,11 +372,11 @@ impl DocumentStore {
     ///         "field_one": "foo",
     ///         "field_two": "bar"
     ///     });
-    ///     let res = doc_store.store(&document.to_string());
+    ///     let res = doc_store.insert(&document.to_string());
     ///     assert!(res.is_ok());
     /// }
     /// ```
-    pub fn store(&self, document: &str) -> Result<(), Box<dyn std::error::Error>> {
+    pub fn insert(&self, document: &str) -> Result<(), Box<dyn std::error::Error>> {
```
prince-chrismc, line 379: It would be nice to have renames like this in separate PRs, this one adds a ton a clutter to the code review.

That being said I do prefer this name.

prince-chrismc (commented):
Some of the diff are so massive they dont render by default.
image

This is a pretty good indication your file needs to be split into smaller chunks
:arrow_right: Please break out the 5 or 6 types into separate files please!

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +176,702 @@ impl Iterator for PackageIterator {
...
+// init_documentStoreName. For example, there is one of these functions for the document store named
+// "package_types". The name of the function is init_package_types. This method creates a Vec of
+// JSON strings that the document store will be pre-populated with.
+//
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Definitions for package types
+const DS_PACKAGE_TYPES: &str = "package_types";
+const IX_PACKAGE_TYPES_NAMES: &str = "names";
+const FLD_PACKAGE_TYPES_NAME: &str = "name";
```
prince-chrismc, line 216: I would really discourage from writing documentation to explain the variable names.

https://github.com/pyrsia/pyrsia/pull/136#discussion_r772736415

```suggestion
const DOCUMENT_STORE_PACKAGE_TYPES: &str = "package_types";
const INDEX_PACKAGE_TYPES_NAMES: &str = "names";
const FIELD_PACKAGE_TYPES_NAME: &str = "name";
```

Please these should not be convoluted and hidden.

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +176,702 @@ impl Iterator for PackageIterator {
...
+/// * `ds_name` ― The name of the document store.
+/// * `index_specs` ― If the document store need to be created, this method is called to get a
+///                   a description of the indexes it will have. When we create a new document store
+///                   we need to define the names of the indexes it will use and the names of the
+///                   fields that each index will be based on.
+/// * `initial_records` ― If the document store needs to be created, this method is called to get
+///                       JSON string that will each be inserted as a record into the document
+///                       store. While most document stores will initially be empty, some need to be
+///                       pre-populated with some records.
+fn open_document_store(
```
prince-chrismc, line 339: https://github.com/pyrsia/pyrsia/pull/136#discussion_r772780018

I must not have been clear. Your answer is not touch on the point I was trying to get across. So let me try again.

:arrow_right: I think most of this logic can be moved to `document_store.rs`.

I suggest having a new function that handles
1. checking if it already exists
2. creating it if it does not exist

I would expect this function to become (more of less -- it's untested pseudo code)

```rust
DocumentStore::open_document_store(name, index_spec);
populate_with_initial_records(&ds, initial_records)?;
Ok(ds)
```

Every single owner of a document store needs to implement the previously mentions bullet points

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +176,702 @@ impl Iterator for PackageIterator {
...
+                "Failed to insert initial record into document store {}: {}\nError was {}",
+                ds.name, record, error
+            );
+            todo!("If an attempt to insert an initial record into document store fails, then we need to do something so that we will know that the document store is missing necessary information")
+        }
+    }
+    Ok(())
+}
+
+fn failed_to_create_document_store(ds_name: &str, error: Box<dyn Error>) -> Result<DocumentStore> {
```
prince-chrismc, line 394: It seems excessive to write a function thats only called one to "format an error"

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -14,26 +14,38 @@
...
+use std::collections::HashMap;
+use std::env;
+use std::error::Error;
+use std::fmt::Debug;
+use std::fs;
 
-// create package version
-
+/// This trait is implemented by structs that provide the MetadataApi.
 pub trait MetadataApi {
```
prince-chrismc, line 38: Other than the `MetadataManager` who else will implement this trait?

I assume the answer is nothing ➡️  Now that the code exists I think we should remove this.

I am not comfortable with the "API" in the name ... it leads me to think this shouldn't exist

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +176,702 @@ impl Iterator for PackageIterator {
...
+
+fn untrusted_metadata_error<'a, T: Signed<'a>>(signed: &T, error: &str) -> anyhow::Result<()> {
+    Err(anyhow!(
+        "New metadata is not trusted: JSON is {}\nError is{}",
+        signed.json().unwrap_or_else(|| "None".to_string()),
+        error
+    ))
+}
+
+// TODO move trust manager trait and its default implementation to a separate module
```
prince-chrismc, line 589: Please move this!

pyrsia_node/src/metadata_manager/metadata.rs
```diff
@@ -172,3 +176,702 @@ impl Iterator for PackageIterator {
...
+fn process_attestations(attestations: Vec<Attestation>) -> Result<()> {
+    debug!(
+        "Found {} valid signatures out of {}",
+        attestations
+            .iter()
+            .filter(|attestation| attestation.signature_is_valid())
+            .count(),
+        attestations.len()
+    );
+    Ok(())
```
prince-chrismc, line 637: Should we fail if there are zero?

prince-chrismc:
Does this address #95  ?

pyrsia_node/src/metadata_manager/metadata.rs
prince-chrismc: https://github.com/rust-lang/api-guidelines/discussions/28#discussioncomment-233414

pyrsia_node/pyrsia_client_lib/src/iso8601.rs
```diff
@@ -0,0 +1,90 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+use log::warn;
+use time::format_description::FormatItem;
+use time::{format_description, OffsetDateTime};
+
+const ISO8601_FORMAT: &str = "[year]-[month]-[day]T[hour]:[minute]:[second].[subsecond digits:3]Z";
+
+fn iso8601_format_spec() -> Vec<FormatItem<'static>> {
```
betarelease, line 22: We should not have our own implementation of Date/Time.

Please use a suitable crate.

betarelease (changes requested):
Let us not add the date time parsing into the main repo. Once we change to using a crate please create an updated PR.

pyrsia_node/pyrsia_client_lib/src/iso8601.rs
mgrand: @betarelease I would happily use the Chrono crate that handles all of ISO8601 parsing and formatting, except that there is a security vulnerability associated with this. According to cargo audit, there is no version available to fix this.
For this reason we have our own. 

You marked as requiring changes. I cannot proceed without your approval now. I know of no crate other than chrono (I did spend quite a bit of time looking) that does what need.  I am unable to proceed until you approve.

prince-chrismc: The CVE comes from the time library which is an optional feature of chrono.
Have you tried changing which feature we enable? to not pull in this deprecate version?

image

From the chrono README

> The oldtime feature is enabled by default for backwards compatibility, but future versions of Chrono are likely to remove the feature entirely.

https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/README.md#L91

https://github.com/chronotope/chrono/blob/f6bd567bb677262645c1fc3131c8c1071cd77ec3/Cargo.toml#L31
https://crates.io/crates/time/0.1.44

mgrand: @betarelease @prince-chrismc The problem is not with being able to represent time. We are using a standard `time` crate for that. The issue is being able to parse and format ISO-8601 dates. That is what we had to implement ourselves.

prince-chrismc: You said https://github.com/pyrsia/pyrsia/pull/143#discussion_r774224940

> I would happily use the Chrono crate that handles all of ISO8601 parsing and formatting, except that there is a security vulnerability associated with this.

So let's try some more solution!

PR #161, opened 2021-12-23
Integrating docker_v2 get_blob to get_artifact of node_manager

### LINKED ISSUE #118: Refactor blobs to Delegate Management of files to the Artifact Manager
Refactor the blobs module to delegate the management of files by calling `node_manager::get_artifact` and `node_manager::put_artifact`.

This will only take care of pulling the artifact , since pushing is after demo. Creating another issue for it.

### LINKED ISSUE #141: extract out docker and node apis to their respective modules
separate out warp apis endpoint from "main" to respective modules

MitaliBo:
closes #118


pyrsia_node/src/docker/v2/handlers/blobs.rs
```diff
@@ -31,7 +34,7 @@ pub async fn handle_get_blobs(
     _name: String,
     hash: String,
 ) -> Result<impl Reply, Rejection> {
-    let blob = format!(
+    /*let blob = format!(
         "/tmp/registry/docker/registry/v2/blobs/sha256/{}/{}/data",
         hash.get(7..9).unwrap(),
         hash.get(7..).unwrap()
```
prince-chrismc, line 40: Does this layout change only impact how we storing the blobs locally?

ie. does the `handle_put_blobs` write in the same place the `get_artifact` will be reading from?

prince-chrismc (commented):
Easy PRs 💟 

One small questions (so I can change my local testing if needs be)

prince-chrismc:
The linked issue is already closed from your last PR 😕 Is that correct?

pyrsia_node/src/docker/v2/handlers/blobs.rs
MitaliBo: Yes to first question and there will be Erwin's changes as he shown today in demo is to pull from dockerhub if not there locally and then store to local artifact_manager. Second is good question but right now we do not full e2e working mechanism for that but the end thought is to have handle_put_blobs will go through artifact_manager to store the blobs at same place based on digest. Also handle_put_blobs does appending of data, so we have to see how it will play through our current implementation of pu_artifact

MitaliBo:
> The linked issue is already closed from your last PR 😕 Is that correct?

tagged wrong PR :(

pyrsia_node/src/docker/v2/handlers/blobs.rs
prince-chrismc: > right now we do not full e2e working mechanism

Well with main today, I can do a `docker push` and then a `docker pull` can get back the same image. So I thought that was working 🤔 

Yep I agree... I saw the put was "in chunks" and to complicate matters @erwin1's demo is also writing data.

I just want to confirm _we are losing some functionality_ until

> the end thought is to have handle_put_blobs will go through artifact_manager to store the blobs at same place based on digest

gets implemented... Is there an issue for this? If yes, may you share it? 🙏

MitaliBo: well i had the code to put blobs based on the artifact_manager's path which takes bytes of data and push it to local registry and then you can pull it from same place, even node_manager's pull and push works for local artifacts, but docker_v2 patch and put calls appending the blobs so i have some questions but since Joeri is on Holidays i have pulled back that part of code. No i will be signing off for holidays in sometime so probably will resume this discussion after we come back and have this PR for draft for now if that makes you comfortable.

prince-chrismc: No I am okay with merging. I just wanted to understand. 

Thanks for the good explanation!

prince-chrismc (approved):

pyrsia_node/src/docker/v2/handlers/blobs.rs
```diff
@@ -56,21 +59,37 @@ pub async fn handle_get_blobs(
...
-    let blob_content = fs::read(blob_path);
-    if blob_content.is_err() {
-        return Err(warp::reject::custom(RegistryError {
-            code: RegistryErrorCode::BlobUnknown,
-        }));
-    }
+    debug!("Getting blob with hash : {:?}", hash);
+
+    let mut art_reader =
+        match get_artifact(&hex::decode(&hash).unwrap().as_ref(), HashAlgorithm::SHA256) {
```
chb0github, line 67: You know, on a result, it has some convenience methods:
https://doc.rust-lang.org/std/result/

In your case:
```rust
art_reader.read_to_end().and_then(|reader| => reader).or_else(|error| Err(...))
```
Then you can save some code

```rust
let good_result: Result<i32, i32> = Ok(10);
let bad_result: Result<i32, i32> = Err(10);

// The `is_ok` and `is_err` methods do what they say.
assert!(good_result.is_ok() && !good_result.is_err());
assert!(bad_result.is_err() && !bad_result.is_ok());

// `map` consumes the `Result` and produces another.
let good_result: Result<i32, i32> = good_result.map(|i| i + 1);
let bad_result: Result<i32, i32> = bad_result.map(|i| i - 1);

// Use `and_then` to continue the computation.
let good_result: Result<bool, i32> = good_result.and_then(|i| Ok(i == 11));

// Use `or_else` to handle the error.
let bad_result: Result<i32, i32> = bad_result.or_else(|i| Ok(i + 20));

// Consume the result and return the contents with `unwrap`.
let final_awesome_result = good_result.unwrap();
```

chb0github (approved):

prince-chrismc:
🎉

PR #164, opened 2021-12-28
Audit security exception for chrono

mgrand:
This is the first step in ripping out a work-around for the chrono security vulnerability.

chrono is cargo.toml but not used. There is be a configuration file for cargo audit that tells it to ignore the chrono issue. There is not any code yet that actually uses chrono.


.github/workflows/rust-lint.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - run: mkdir -p ~/.cargo
```
efrisch, line 14: If $HOME doesn't exist, `-p` might not create it. I don't think it hurts anything though.

.github/workflows/rust-lint.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
```
efrisch, line 11: We should probably pin a specific ubuntu version; else this could break when ubuntu has a new release.

.github/workflows/rust-lint.yml
```diff
@@ -11,6 +11,8 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - run: mkdir -p ~/.cargo
+      - run: cp .github/workflows/.cargo/* ~/.cargo
```
efrisch, line 15: - run: (cd $HOME/.github/workflows/.cargo; tar cfp - .) | (cd $HOME/.cargo; tar xf -)

efrisch (approved):
👍

.github/workflows/rust-lint.yml
mgrand: @efrisch Thank you. I will use $HOME instead of ~

.github/workflows/rust-lint.yml
mgrand: @efrisch I will write an issue for us to use a specific Ubuntu version.

prince-chrismc: We should not need to.

1. This is an open source project we need to work regardless of platform
2. We have months of lead time before latest points to the next release

.github/workflows/rust-lint.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - run: mkdir -p $HOME/.cargo
+      - run: cp .github/workflows/.cargo/* $HOME/.cargo
```
prince-chrismc, line 15: There are a lot of files read from this directory. I am not sure a glob star is the best.

Lets be explicit so we dont have surprises.

```suggestion
      - run: cp .github/workflows/.cargo/audit.toml $HOME/.cargo
```

.github/workflows/.cargo/audit.toml
```diff
@@ -0,0 +1,5 @@
+# All of the options which can be passed via CLI arguments can also be permanently specified in this file.
+# A good example is at https://docs.rs/crate/cargo-audit/0.10.0/source/audit.toml.example
+
+[advisories]
+ignore = ["RUSTSEC-2020-0071", "RUSTSEC-2020-0159"] # advisory IDs to ignore
```
prince-chrismc, line 5: We should not ignore the vulnerability with time since that has a fix.

```suggestion
ignore = ["RUSTSEC-2020-0159"] # advisory IDs to ignore
```

pyrsia_node/Cargo.toml
```diff
@@ -11,6 +11,7 @@ anyhow = "1.0.45"
 bincode = "1.3.3"
 blake3 = "1.2.0"
 bytes = "1.1.0"
+chrono = "0.4.19"
```
prince-chrismc, line 14: Lets not require old time! https://github.com/chronotope/chrono/blob/f6bd567bb677262645c1fc3131c8c1071cd77ec3/Cargo.toml#L27 

```suggestion
chrono = { verion="0.4.19", default= false, features=["clock", "std"] }
```

prince-chrismc (changes requested):
We should aim to introduce the least amount of exceptions.

Given the comments in slack, the team has the same feeling, so I am requesting changes to make sure we address the concerns.

.github/workflows/.cargo/audit.toml
mgrand: Including cargo raises both alerts. What is the fix for RUSTSEC-2020-0071

pyrsia_node/Cargo.toml
mgrand: The suggested fix did not work. See https://github.com/pyrsia/pyrsia/runs/4653072658?check_suite_focus=true

I will leave this change in and put the second ignore back in.

mgrand: @prince-chrismc I tried it again with cargo.lock updated. It still does not work.  Apparently cargo audit isn't as smart as you are.  See
https://github.com/pyrsia/pyrsia/runs/4653351944?check_suite_focus=true

efrisch (dismissed):
👍

prince-chrismc:
Please consider https://github.com/mgrand/pyrsia/pull/1 against your branch 🙏

prince-chrismc (approved):

efrisch (approved):
👍

PR #168, opened 2021-12-28
Disable complaints about generated new methods with too many arguments.

mgrand:
Clippy is complaining that the mechanically generated `new` methods generated by the `#[signed_struct]` macro have too many parameters if the struct in question has more than 7 fields.

I have modified the macro to generate a 
#[allow(clippy::too_many_arguments)]
before the `new` methods to stop the complaining


prince-chrismc (approved):

fishseabowl:
That's fine for me. Alternatively, could we change too-many-arguments-threshold to 14 or make a simpler signed_struct?

mgrand:
@fishseabowl You reviewed without approving. Are you saying that the structs that we sign should not contain more than some maximum number of fields? It is not clear to me what you think I should do before you approve.

fishseabowl (approved):

PR #169, opened 2021-12-28
Added gossipsub file-transfer topic for magnet links

efrisch:


pyrsia_node/src/main.rs
```diff
@@ -114,9 +114,10 @@ async fn main() {
                                                                             //let floodsub_topic: Topic = floodsub::Topic::new("pyrsia-node-converstation"); // Create a Floodsub topic
     let (respond_tx, respond_rx) = mpsc::channel(32);
     let floodsub_topic: Topic = floodsub::Topic::new("pyrsia-node-converstation");
+    let gossip_topic: libp2p::gossipsub::IdentTopic = libp2p::gossipsub::IdentTopic::new("pyrsia-file-share-topic");
```
mgrand: This is getting to be a very long method. Perhaps you could move your code to a separate method or function.

pyrsia_node/src/main.rs
```diff
@@ -114,9 +114,10 @@ async fn main() {
...
     // Create a Swarm to manage peers and events.
     let mut swarm: MyBehaviourSwarm =
-        new_swarm(floodsub_topic.clone(), transport, local_peer_id, respond_tx)
+        new_swarm(gossip_topic.clone(), floodsub_topic.clone(), transport, local_key, respond_tx)
```
mgrand: The usual way to reference this in Rust would be `swarm::new`. That is the Rust convention. Because it was called new_swarm, I thought it was defined locally until I found the `use`.

pyrsia_node/src/main.rs
```diff
@@ -202,6 +203,9 @@ async fn main() {
                 }
                 EventType::Input(line) => match line.as_str() {
                     "peers" => swarm.behaviour_mut().list_peers_cmd().await,
+                    cmd if cmd.starts_with("magnet:") => {
+                        println!("{}", swarm.behaviour_mut().gossipsub_mut().publish(gossip_topic.clone(), cmd).unwrap())
```
mgrand: From within the pyrsia node, we should be using the log crate for this sort of thing. Perhaps instead of 
`println!`
use
`info!`

pyrsia_node/src/network/behavior.rs
```diff
@@ -90,6 +98,30 @@ impl MyBehaviour {
...
+impl NetworkBehaviourEventProcess<gossipsub::GossipsubEvent> for MyBehaviour {
+    // Called when `gossipsub` produces an event.
+    fn inject_event(&mut self, message: gossipsub::GossipsubEvent) {
+        if let gossipsub::GossipsubEvent::Message {
+            propagation_source,
+            message_id,
+            message,
+        } = message
+        {
+            let msg_data: String = String::from_utf8(message.data).unwrap();
```
mgrand, line 110: If unwrap fails, it panics. If message.data is not valid UTF8, I think it would be better to log an error and return.

pyrsia_node/src/network/swarm.rs
```diff
@@ -16,31 +16,71 @@
...
+use libp2p::{gossipsub, identity, swarm::SwarmEvent, Multiaddr};
+use std::collections::hash_map::DefaultHasher;
+use std::error::Error;
+use std::hash::{Hash, Hasher};
+use std::time::Duration;
 
 pub type MyBehaviourSwarm = Swarm<MyBehaviour>;
 
 pub async fn new(
+    gossip_topic: IdentTopic,
```
mgrand, line 39: This is large enough to break up into smaller methods.

pyrsia_node/src/network/swarm.rs
```diff
@@ -16,31 +16,71 @@
...
+        let mut s = DefaultHasher::new();
+        message.data.hash(&mut s);
+        MessageId::from(s.finish().to_string())
+    };
+
+    // Set a custom gossipsub
+    let gossipsub_config = gossipsub::GossipsubConfigBuilder::default()
+        .heartbeat_interval(Duration::from_secs(10)) // This is set to aid debugging by not cluttering the log space
+        .validation_mode(ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
+        .message_id_fn(message_id_fn) // content-address messages. No two messages of the
```
mgrand, line 61: shouldn't you be passing the message to the function rather than itself?

mgrand (commented):
I left some comments.

mgrand (approved):

pyrsia_node/src/network/swarm.rs
```diff
@@ -16,31 +16,71 @@
...
+use std::time::Duration;
 
 pub type MyBehaviourSwarm = Swarm<MyBehaviour>;
 
 pub async fn new(
+    gossip_topic: IdentTopic,
     topic: Topic,
     transport: TcpTokioTransport,
-    peer_id: PeerId,
+    local_key: identity::Keypair,
```
prince-chrismc, line 42: Why are there no change in main? I'd expect the main to low it's usage of `PeerId` 

❓ Are we running with two ids?
❓ Whats the old one doing (in main)?

tests/test_code_coverage.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-docker build --no-cache -t code_coverage:2.0 -f ./tests/code_coverage.Dockerfile .
-docker run --rm --security-opt seccomp=unconfined -it code_coverage:2.0
+docker build --no-cache -t code_coverage:1.0 -f ./tests/code_coverage.Dockerfile .
+docker run --rm --security-opt seccomp=unconfined -it code_coverage:1.0
```
prince-chrismc, line 4: Looks like merge conflicts

```suggestion
docker build --no-cache -t code_coverage:2.0 -f ./tests/code_coverage.Dockerfile .
docker run --rm --security-opt seccomp=unconfined -it code_coverage:2.0
```

pyrsia_node/src/network/behavior.rs
```diff
@@ -90,6 +98,30 @@ impl MyBehaviour {
...
+            if msg_data.starts_with("magnet:") {
+                // Synapse RPC Integration point
+                println!("Start downloading {}", msg_data);
```
prince-chrismc, line 113: node should not be printing to console =)

```suggestion
                info!("Start downloading {}", msg_data);
```

pyrsia_node/src/main.rs
```diff
@@ -202,6 +209,16 @@ async fn main() {
                 }
                 EventType::Input(line) => match line.as_str() {
                     "peers" => swarm.behaviour_mut().list_peers_cmd().await,
+                    cmd if cmd.starts_with("magnet:") => {
+                        println!(
```
prince-chrismc, line 213:
```suggestion
                        info!(
```

pyrsia_node/src/network/swarm.rs
prince-chrismc: Might be out of scope but I'd of expected this to be a member I guess... It's error prone to "extract" the id in multiple places.

https://github.com/pyrsia/pyrsia/blob/44022d8d429c8b0d31a02c42dd37e612e9ffcba2/pyrsia_node/src/main.rs#L229

prince-chrismc (commented):
small nits about printing to console

prince-chrismc:
Please sign the CLA

pyrsia_node/src/network/swarm.rs
efrisch: We're using the same peer_id with two topics. One is floodsub, one is gossipsub; but both share the public key. Which is what this is doing. Open to changes if there's a better approach. We can discuss.

tests/test_code_coverage.sh
efrisch: I deliberated changed this back to version 1.0; I was having an issue on my machine with 2.0 - and we explicitly use `--rm` with docker run so the version should never increment.

pyrsia_node/src/network/behavior.rs
efrisch: This was a placeholder. But a logger statement serves as well. Changed.

pyrsia_node/src/network/swarm.rs
efrisch: That's a message id function (not a message itself); it gets invoked to generate a message id on every message and ensures uniqueness.

pyrsia_node/src/main.rs
efrisch: Also changed.

efrisch:
> Please sign the CLA

Done. Thought I was covered by contractual relationship with jfrog. Easy enough to be explicit about it.

mgrand (dismissed):

mgrand (dismissed):

.github/workflows/cla.yml
```diff
@@ -0,0 +1,35 @@
+name: "CLA Assistant"
```
prince-chrismc, line 1: What is this?

prince-chrismc (changes requested):

prince-chrismc:
Please delete the new workflow. You must have been out sick when I presented it to the team.

You (and any future contributors) only need to follow the comment that is posted in our PR
https://github.com/pyrsia/pyrsia/pull/169#issuecomment-1002300469 for example.

tests/test_code_coverage.sh
prince-chrismc: That not what `--rm` does 😕 

"remove" only cleans up the container not the image that is created.

What issue did you have? when creating an image you can give it any name. There's likely a bigger issue if you can not change the name.

pyrsia_node/src/network/swarm.rs
prince-chrismc: Let me make this a new issue.

error prone code == bad

tests/test_code_coverage.sh
efrisch: code_coverage is the name 2.0 is a tag. 2.0 implies there was a 1.0 tag that isn't published. I guess I want to know why we would skip 1.0 and jump to 2.0? If you really feel strongly about it I'll investigate further into the issue I had.

From `docker build --help`, `-t, --tag list                Name and optionally a tag in the 'name:tag' format`

.github/workflows/cla.yml
efrisch: https://jfrog.slack.com/archives/C02GCU7PE8L/p1639173359470100?thread_ts=1639173356.470000&cid=C02GCU7PE8L

tests/test_code_coverage.sh
prince-chrismc: We will never publish this. It's just a local test image.
I changed the version because I changed the internal structure and broke the API on the image.

Let's remove version (and the confusion) seems like its worse to have it

```suggestion
docker build --no-cache -t code_coverage -f ./tests/code_coverage.Dockerfile .
docker run --rm --security-opt seccomp=unconfined -it code_coverage
```

.github/workflows/cla.yml
prince-chrismc: But if you read the thread... I said we should not do that. I also share a video demonstrating the working CLA 😕

pyrsia_node/src/network/swarm.rs
```diff
@@ -16,31 +16,71 @@
...
 ) -> Result<MyBehaviourSwarm, ()> {
+    let local_peer_id = PeerId::from(local_key.public());
     //create kad
-    let store = MemoryStore::new(peer_id);
-    let kademlia = Kademlia::new(peer_id, store);
+    let store = MemoryStore::new(local_peer_id);
+    let kademlia = Kademlia::new(local_peer_id, store);
+
+    // To content-address message, we can take the hash of message and use it as an ID.
+    let message_id_fn = |message: &GossipsubMessage| {
```
chb0github, line 51: This closure doesn't have anything it's capturing - it could be a top-level function

tests/test_code_coverage.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-docker build --no-cache -t code_coverage:2.0 -f ./tests/code_coverage.Dockerfile .
-docker run --rm --security-opt seccomp=unconfined -it code_coverage:2.0
+docker build --no-cache -t code_coverage -f ./tests/code_coverage.Dockerfile .
```
chb0github, line 3: We could achieve this with a `docker-compose.yaml` file.

chb0github (approved):

prince-chrismc (approved):

prince-chrismc:
Seems like a random failure?

https://github.com/pyrsia/pyrsia/runs/4661021813?check_suite_focus=true#step:5:409

"happy path run for more than 60s" 😖

PR #170, opened 2021-12-28
Use chrono crate for data time

mgrand:
In a previous PR we disabled the security alert caused by a vulnerability in chrono. In this PR we remove the homegrown code to parse and format IOS8601 dates and use chrono for that instead.


prince-chrismc (approved):

efrisch (approved):
👍

Issue #162, opened 2021-12-23
Capture Manifests on the node.


mgrand:
Capture docker manifests as artifacts, create metadata that identifies manifest artifacts and allow the manifests to be retrieved by the docker integration and shared with the Pyrsia network. This will involve these subtasks:

- [x] #171
- [x] #176
- [x] #212
- [x] #213
- [ ] #214
- [ ] #215

This is nice to have but not required for demo

prince-chrismc:
I was under the impression this was already supported? 🤔 

https://github.com/pyrsia/pyrsia/blob/fbe7fb1b6e09adaaab9e7df90c3d2311f2e9915a/pyrsia_node/src/docker/v2/handlers/manifests.rs#L68

Perhaps "capture" does not mean working/saving? That's my understanding

erwin1:
I created #268 to store docker manifests in the artifact manager when we do a pull-through of docker images.

betarelease:
Closing this since #268 should handle what is relevant and for other things we should have more current thinking and create new tickets.

Issue #114, opened 2021-12-16
Docker integration response hardcode some information


prince-chrismc:
Playing with the implementation I noticed a small limitation

- hostname
- port

prince-chrismc:
https://github.com/pyrsia/pyrsia/blob/35ddbab30b2e50c66390ba9e0f8bb88aca196273/src/docker/v2/handlers/blobs.rs#L96

Still true 🤔

betarelease:
Closing because it is outdated and code has evolved.

PR #174, opened 2021-12-29
Add unit test to manifest.rs

mgrand:


pyrsia_node/src/docker/v2/handlers/manifests.rs
```diff
@@ -175,7 +179,7 @@ pub async fn handle_put_manifest(
 
         Ok(warp::http::response::Builder::new()
             .header(
-                "Location",
+                LOCATION,
```
chb0github, line 182: Why?

pyrsia_node/src/docker/v2/handlers/manifests.rs
```diff
@@ -187,3 +191,86 @@ pub async fn handle_put_manifest(
...
+            )
+            .await
+        };
+        let result = executor::block_on(future);
+        match result {
+            Ok(reply) => {
+                let response = reply.into_response();
+                assert_eq!(response.status(), 201);
+                assert!(response.headers().contains_key(LOCATION));
+                let location = response.headers().get(LOCATION).unwrap().to_str()?;
```
chb0github, line 268: You're not doing anything with this variable

pyrsia_node/src/docker/v2/handlers/manifests.rs
```diff
@@ -187,3 +191,86 @@ pub async fn handle_put_manifest(
...
+            }
+            Err(rejection) => {
+                assert!(false)
```
chb0github, line 271: isn't there something like `fail` instead of `assert(false)`

pyrsia_node/src/docker/v2/handlers/manifests.rs
```diff
@@ -187,3 +191,86 @@ pub async fn handle_put_manifest(
...
+                Bytes::from(MANIFEST_JSON.as_bytes()),
+            )
+            .await
+        };
+        let result = executor::block_on(future);
+        match result {
+            Ok(reply) => {
+                let response = reply.into_response();
+                assert_eq!(response.status(), 201);
+                assert!(response.headers().contains_key(LOCATION));
```
chb0github, line 267: So, as long as the header exists, the value doesn't matter?

chb0github (dismissed):
Only suggestions

prince-chrismc:
Please link the related issue 🙏 

image

pyrsia_node/src/docker/v2/handlers/manifests.rs
prince-chrismc: Maybe at the top ? No matching?

```
assert!(reply.is_ok());
```

pyrsia_node/src/docker/v2/handlers/manifests.rs
prince-chrismc: It does but it's already bad see #114... so it might be out of scope for now

pyrsia_node/src/docker/v2/handlers/manifests.rs
mgrand: Thank you for pointing that out.

pyrsia_node/src/docker/v2/handlers/manifests.rs
mgrand: There is nothing like `fail` as far as I can tell.
Neither is there an is_ok.  Matching OK is the way to know that the response was a success response. That is just the way that the warp framework works.

pyrsia_node/src/docker/v2/handlers/manifests.rs
mgrand: Before I am finished with [#162](https://github.com/pyrsia/pyrsia/issues/162) this should work

chb0github (approved):

efrisch (approved):
👍

PR #175, opened 2021-12-30
Put docker manifests in artifact manager

mgrand:
manifests.rs now puts manifests in the artifact manager.

### DIFF
--- a/pyrsia_node/src/node_manager/handlers.rs
+++ b/pyrsia_node/src/node_manager/handlers.rs
@@ -30,7 +30,7 @@ use std::str;
 const ART_MGR_DIR: &str = "pyrsia";
 
 lazy_static! {
-    static ref ART_MGR: ArtifactManager = {
+    pub static ref ART_MGR: ArtifactManager = {
         fs::create_dir_all(ART_MGR_DIR).expect("Error creating dir for artifacts");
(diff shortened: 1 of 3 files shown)


chb0github (approved):

pyrsia_node/src/node_manager/handlers.rs
```diff
@@ -30,7 +30,7 @@ use std::str;
 const ART_MGR_DIR: &str = "pyrsia";
 
 lazy_static! {
-    static ref ART_MGR: ArtifactManager = {
+    pub static ref ART_MGR: ArtifactManager = {
```
efrisch, line 33: This looks suspiciously like a Singleton. `ART_MGR_INSTANCE`?

efrisch (dismissed):
Looks reasonable to me. 👍

pyrsia_node/src/node_manager/handlers.rs
mgrand: This is a statically allocated instance of an ArtifactManager. Access to this is not managed by an object, so technically it is not a singleton. This made it through the review process  before. All I did was make this public.

efrisch (approved):
👍

fishseabowl (approved):
this seems reasonable to me.

PR #179, opened 2021-12-30
Add new block definition for discuss

fishseabowl:
I would add methods and traits later. Welcome any suggestions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -44,3 +44,18 @@ jobs:
         with:
           path: .
           config: .github/license-check/config.json
+
+  inclusive-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: get-woke/woke-action-reviewdog@v0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
+          reporter: github-pr-review
+          # Change reporter level if you need.
+          # GitHub Status Check won't become failure with warning.
+          level: warning
+          # Enable this to fail the check when violations are found
+          # fail-on-error: true


pyrsia_node/src/block_chain/block.rs
```diff
@@ -24,3 +25,67 @@ pub struct Block {
...
+    pub committer: Address,      //Beneficiary Address, currently this is the commit node PeerID
+    pub transactions_root: H256, //256bit Keccak Hash of the root node of Transaction Trie
+    //    pub state_root: H256,   //256bit Keccak Hash of the root node of state Trie
+    pub timestamp: u64,
+    pub number: U256,
+    pub nonce: U256,
+    pub signature: BlockSignature,
+}
+
+pub struct block {
```
chb0github, line 44: Upper case please

pyrsia_node/src/block_chain/block.rs
```diff
@@ -24,3 +25,67 @@ pub struct Block {
...
+
+pub struct Transaction {
+    pub nonce: U256,
+    pub trans_type: TransactionType,
+
+    pub signature: TransactionSignature,
+    pub payload: Vec<u8>,
+}
+
+pub struct TransactionSignature {
```
chb0github, line 57: How about we just call these `signature`? They are identical.

Also, I was thinking that the signature should include a crypto challenge

chb0github (changes requested):
see comments

pyrsia_node/src/block_chain/block.rs
mgrand: @fishseabowl @chb0github The data that we store will be JSON that has its own signature. See the `Signed` trait in `pyrsia_node/pyrsia_client_lib/src/signed.rs`.  I would like to see the blockchain rely on the signature in the JSON.

The signers of the JSON will be signing their JSON outside of the node. It is unclear who would add another signature when we store the JSON in the blockchain.   

I think it will be safe to assume that all data stored on the blockchain will be a struct that implements the `Signed` trait. This means that the struct has a signed JSON object associated with it. The trait includes methods for verifying the signature and for getting the JSON.

For examples see the unit tests that the end of `pyrsia_node/pyrsia_client_lib/src/signed.rs` and also in 
`pyrsia_node/pyrsia_client_lib/signed_struct/tests/lib.rs`

The first type of data that we will want to store on the block chain is an identity. These will be instances of the `Identity` struct at `pyrsia_node/src/node_manager/model/identity.rs`.

The `Identity` struct has a `public_key` field which will serve as the primary key for the identity. We will need to be able to quickly retrieve identities by their public key. We will also need to be able to update the information in an identity, subject to signature constraints. We will eventually also need to be able to retrieve historical identity records rather than just the current identity information.

mgrand (dismissed):

pyrsia_node/src/block_chain/block.rs
fishseabowl: Good comments. Thanks.

pyrsia_node/src/block_chain/block.rs
fishseabowl: > How about we just call these `signature`? They are identical.
> 
> Also, I was thinking that the signature should include a crypto challenge

Yes. Maybe they are identical, maybe not. I would create Signature struct and also have two alias name TransactionSignature and BlockSignature.

fishseabowl: @mgrand Thanks for your comments. 1. I just think identity is only part of business logic, so these could be stored in transaction.payload. 2. Converting between different data formats interface would be provided.

fishseabowl: @mgrand I am not sure your goal looks like DID(Decentralized identifiers). Maybe hyperledger indy is a good reference.

mgrand: I am not familiar with Hyperledger Indy. I will take a look. Thank you for mentioning it

mgrand: Even though transactions are signed, blocks will also need to be signed. The originator of the transaction or signed struct will sign that. The committing node will need to sign the block. 

The reason that we need the committer to sign the block is that the blocks will be propagated to non-commiter and non-consensus node using an insecure mechanism. Signing the blocks allows us to detect any attempts to introduce bogus blocks into the Pyrsia network.

mgrand: @fishseabowl I took a quick look at Indy. It is interesting, but not what we need. I see two areas where its goals are different than ours:
* It assumes that all identity providers provide information using the same interoperable format. We need to accommodate diverse and heterogeneous sources of identity information.
* It features, "Zero Knowledge Proofs which prove that some or all of the data in a set of Claims is true without revealing any additional information, including the identity of the Prover"  Because we are trying to create a process that automates trust, we need everything about an identity claim to be transparent.

fishseabowl: @mgrand Thanks for your comment. Hyperledger Indy have some good features: 1. Delegate 2. Revoke. It may be helpful.  Otherwise, I think ZKP just want to leverage transparent and privacy.

pyrsia_node/src/block_chain/block.rs
```diff
@@ -24,3 +25,67 @@ pub struct Block {
     pub data: String,
     pub nonce: u64,
 }
+
+pub struct H256(pub [u8; 32]); //Unformatted binary data of fixed length
+pub struct Address(pub [u8; 20]); //Unformatted binary data of fixed length
+pub struct U256(pub [u64; 4]); //Little-endian large integer type
+
+pub struct Header {
```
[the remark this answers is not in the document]
prince-chrismc: For this to be stored as JSON we could use you help answering a few questions (please open a PR showing how to complete the following)

- Serialize custom type (e.g `Address`)
  - Is this a custom type or just an alias
  - Are arrays of ints going to be `[ 1, 2, 2]`
  - How could we just a HTTP safe string thats base64 encoded from this?
  - How can we get a hex encoded version of this value?

This would be a great issue for you to open and pick up ❤️

pyrsia_node/src/block_chain/block.rs
```diff
@@ -24,3 +25,67 @@ pub struct Block {
...
+    pub committer: Address,      //Beneficiary Address, currently this is the commit node PeerID
+    pub transactions_root: H256, //256bit Keccak Hash of the root node of Transaction Trie
+    //    pub state_root: H256,   //256bit Keccak Hash of the root node of state Trie
+    pub timestamp: u64,
+    pub number: U256,
+    pub nonce: U256,
+    pub signature: BlockSignature,
+}
+
+pub struct Block_v1 {
```
chb0github, line 44: Just replace `Block` - we don't need a `v....` - I realize this would mean the app won't compile, but if we are going to use your structs, then it's going to have to all be back ported anyways.

Please just overwrite `Block` and call it done

prince-chrismc:
You need to fix the commit authorship for the CLA. See https://github.com/pyrsia/pyrsia/pull/179#issuecomment-1003210392

prince-chrismc:
Looking over this, I agree the structure is a great step. Please continue with https://github.com/pyrsia/pyrsia/pull/179#discussion_r778288955 and updating the rest of the code (or leave a few todos)

prince-chrismc:
I think the signing is out of scope for this PR. Let's get something working.

chb0github:
My assumption is that this code/struct would be integrated into the code base. Even if you have to rip out everything and just stub it to make it compile. We need to move forward

fishseabowl:
> I think the signing is out of scope for this PR. Let's get something working.

At first, I would directly use libp2p::identity::ed25519::Keypair to sign and verify.

chb0github:
Xuefe, will you be merging this in?

prince-chrismc:
Moving this as draft since there are lot of merge conflicts 😱 this will give you time to fix it 👍

prince-chrismc:
replaced by #192 🤔

Issue #148, opened 2021-12-22
oepnssf: Inclusivity and Inclusive Naming Statement


sbtaylor15:
Need to be inclusive in our documentation, meetings and coding.  This goes beyond the code of conduct.  The code of conduct covers when folks are "bad actors" and how to handle them.  Inclusivity is needed to make everyone feel welcome to the project.

A `values.md` is needed to state our Inclusivity Goals and how to archive those goals by following Inclusive Naming in our documentation, meeting minutes, and code.  We all use non-inclusive terms unintentionally and out of habit.  Our `values.md` should state that we strive to resolved these issues as fast as possible.   

Recommend we follow: https://inclusivenaming.org/

PR #187, opened 2022-01-04
Enable Inclusive Linting

sbtaylor15:
Add linting to ensure we a conforming to inclusive naming guidelines.

See [Inclusive Linting](https://microsoft.github.io/code-with-engineering-playbook/continuous-integration/inclusive-linting/) for details on the action.


sbtaylor15:
@prince-chrismc - getting this error with the token.   Not sure how to resolve it.  The action is trying to add a comment to the PR.

```
Running woke with reviewdog 🐶 ...
  reviewdog: This GitHub token doesn't have write permission of Review API [1], 
  so reviewdog will report results via logging command [2] and create annotations similar to
  github-pr-check reporter as a fallback.
```

prince-chrismc:
Sounds like the usually lack of permission towards public forks.

- does this action download any code? (e.g `npm install` would)
- does this PR run any code from the repo? (e.g `cargo test`)

sbtaylor15:
The action downloads a golang program to scan the repo for "bad" words.  It does not run the application.
[woke](https://github.com/get-woke/woke) is the program being run.  Then reviewdog is used to make updates to the PR via comments.

prince-chrismc:
Actually looking into the logs, seem like the author actions already made it smart! 🚀 https://github.com/pyrsia/pyrsia/runs/4707393792?check_suite_focus=true#step:3:32

I would much prefer to have this under the "lint" workflow... we can remove the rust name.
The "checks api" from public forks is quite annoying, it leave reports under the wrong action. I would suggest using the https://github.com/get-woke/woke-action-reviewdog#github-pr-review reporter along with the the `"pull_request_target"` it seems safe at first glance 🤞

sbtaylor15:
@prince-chrismc - Is this what you are thinking of for the structure?

prince-chrismc:
Yes ❤️ 

I think we will have permission problems however. Maybe we can merge this an follow up with more adjustments laster?

prince-chrismc:
Does this close #148?

prince-chrismc (approved):

sbtaylor15:
> Does this close #148?

No, #148 is a "values.md" that says how we approach inclusivity.  That file still needs to be created.

mgrand (approved):

prince-chrismc:
This is is never being ran 😕 

image

PR #190, opened 2022-01-04
Blockchain refinement

prince-chrismc:
This is the work we got lost in before the holidays.

Closing issue since they are no longer relevant:
- We tried to define traits but that was unsuccessful.
- We tried to connect the block chain code to the node api and that proved too difficult.

We are looking to setup a new approach.


pyrsia_node/src/block_chain/block.rs
```diff
@@ -24,3 +26,27 @@ pub struct Block {
     pub data: String,
     pub nonce: u64,
 }
+
+impl Display for Block {
```
chb0github, line 30: @fishseabowl - please roll in these implementations for block into your code.

prince-chrismc: They'll get picked up if we both merge 🤔

pyrsia_node/src/block_chain/block.rs
```diff
@@ -24,3 +26,27 @@ pub struct Block {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let json = serde_json::to_string_pretty(&self).expect("json format error");
+        write!(f, "{}", json)
+    }
+}
+/*
+impl std::fmt::Display for HashAlgorithm {
```
chb0github, line 37: ditch

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -29,35 +31,26 @@ const DIFFICULTY_PREFIX: &str = "00";
...
+                data: String::from("genesis!"),
+                nonce: 2836,
+                hash: "0000f816a87f806bb0073dcf026a64fb40c946b5abee2573702828694d5b4c43"
+                    .to_string(),
+            }]),
+        }
     }
-    pub(crate) fn dump(&self) -> Result<String, Error> {
+
+    pub fn dump(&self) -> Result<String, Error> {
```
chb0github, line 47: This can probably be removed and we can just use `format`

pyrsia_node/src/block_chain/block_chain.rs
```diff
@@ -76,45 +69,50 @@ impl BlockChain {
...
     }
 
     fn is_chain_valid(&self, chain: &[Block]) -> bool {
         for i in 1..chain.len() {
             let first = chain.get(i - 1).expect("has to exist");
             let second = chain.get(i).expect("has to exist");
-            if !self.is_block_valid(second, first) {
-                return false;
-            }
+            match BlockChain::are_blocks_sequential(first, second.clone()) {
```
chb0github, line 108: This really should return `Result<Block,Err>` and be renamed. But, not essential

chb0github:
I mean, I guess we can add it. But it just seems like little value add

prince-chrismc:
> it just seems like little value add

Thats better then nothing! Plus this way it saves what we tried. I am going to tidy it up some more so we can merge soon

pyrsia_node/src/block_chain/block.rs
fishseabowl: ok.

pyrsia_node/src/block_chain/block_chain.rs
prince-chrismc: I did not understand this comment, so I skipped it for now 🙈

pyrsia_node/src/main.rs
```diff
@@ -139,18 +140,26 @@ async fn main() {
...
     let my_stats = shared_stats.clone();
     let tx2 = tx.clone();
 
-    let my_stats1 = shared_stats.clone();
-    let tx3 = tx.clone();
+    // We need to have two channels (to seperate the handling)
+    // 1. API to main
+    // 2. main to API
+    let (blocks_get_tx_to_main, mut blocks_get_rx) = mpsc::channel(32); // Request Channel
+    let (blocks_get_tx_answer, blocks_get_rx_answers_from_main) = mpsc::channel(32); // Response Channel
```
fishseabowl, line 155: Can we name these variants more clearly, for example: main_send, main_rec, chain_send, chain_rec?

prince-chrismc:
Replace multiple directional channels with lazy static block chain

It's painfully obvious the two channels for communicating back and forth with an object is just too complicated at this stage so we should use the other pattern
https://github.com/pyrsia/pyrsia/blob/e83f4f4f79e41507229b2f5ae401844d690cabfe/src/node_manager/handlers.rs#L32-L37
which is much easier to swallow

***

Ill make this an issue after it's merged

fishseabowl (approved):
That's fine with me.

chb0github (approved):

PR #196, opened 2022-01-06
Docker registry pull through

### LINKED ISSUE #123: The Pyrsia node pulls in images from Docker Hub when needed, and makes them available in the Pyrsia network
This should happen when an image is requested throught the registry API:
* try to find the image in Pyrsia local storage
* then try to find the image in the Pyrsia network
* if still not found, pull the image from Docker Hub, store locally and publish in the Pyrsia network

tiainen:
Fixes #123


pyrsia_node/src/docker/v2/handlers/blobs.rs
```diff
@@ -39,17 +40,19 @@ pub async fn handle_get_blobs(
...
-
-        match tx.send(send_message).await {
-            Ok(_) => debug!("hash sent"),
-            Err(_) => error!("failed to send stdin input"),
+        if let Err(_) = get_blob_from_docker_hub(&_name, &hash).await {
+            let mut send_message: String = "get_blobs | ".to_owned();
+            let hash_clone: String = hash.clone();
+            send_message.push_str(&hash_clone);
+
+            match tx.send(send_message).await {
```
prince-chrismc, line 48: Should we not look on the pyrsia network before talking to docker hub?

(might be a follow up PR, former wasnt finished, but the order looks backwards)

pyrsia_node/src/docker/v2/handlers/blobs.rs
```diff
@@ -137,88 +123,115 @@ pub async fn handle_put_blob(
...
-        return Err(e);
-    }
+async fn get_blob_from_docker_hub(name: &str, hash: &str) -> Result<String, Rejection> {
+    let token = get_docker_hub_auth_token(name).await?;
 
-    Ok((initial_file_length, total_bytes_read))
+    get_blob_from_docker_hub_with_token(name, hash, token).await
+}
+
+async fn get_blob_from_docker_hub_with_token(
```
prince-chrismc, line 209: I looked quickly and saw this gem https://docs.rs/oauth2/latest/oauth2/

Maybe next PR we can use a crate?

pyrsia_node/src/docker/v2/handlers/manifests.rs
```diff
@@ -109,120 +105,135 @@ pub async fn handle_put_manifest(
...
+    Ok(hash)
+}
+
+async fn get_manifest_from_docker_hub(name: &str, tag: &str) -> Result<String, Rejection> {
+    let token = get_docker_hub_auth_token(name).await?;
+
+    get_manifest_from_docker_hub_with_token(name, tag, token).await
+}
+
+async fn get_manifest_from_docker_hub_with_token(
```
prince-chrismc, line 205: again using a crate could make this more DRY https://docs.rs/oauth2/latest/oauth2/

prince-chrismc (commented):
One and and a few questions/future ideas... otherwise top notch like always 🚀

pyrsia_node/src/docker/v2/handlers/blobs.rs
johanvos: True, that is what should be happening (looking on the pyrsia network) -- and that is indeed a follow-up PR.

prince-chrismc:
I noticed this switched from artifact manager to local storage, is that part of the plan with @MitaliBo ?

MitaliBo:
> I noticed this switched from artifact manager to local storage, is that part of the plan with @MitaliBo ?

yes so that i can work upon their changes to integrate with artifact_manager e2e rather half here and half there and since Joeri and team are not much upto track with artifact_manager also

prince-chrismc:
Perfect, thanks for filling in 🤗

prince-chrismc (dismissed):

pyrsia_node/src/docker.rs
```diff
@@ -14,5 +14,6 @@
    limitations under the License.
 */
 
+mod docker_hub_util;
```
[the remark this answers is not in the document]
tiainen: The module needs to be declared so the rust compiler is able to find it. I've left it private as it doesn't contain anything useful outside of the pyrsia project.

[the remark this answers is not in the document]
prince-chrismc: Thanks for confirming!

MitaliBo (approved):

src/docker/v2/handlers/blobs.rs
```diff
@@ -82,7 +83,7 @@ pub async fn handle_get_blobs(
     send_message.push_str(&hash_clone);
     tx.send(send_message.clone());
 
-    debug!("Getting blob with hash : {:?}", hash);
+    /*     debug!("Getting blob with hash : {:?}", hash);
```
mgrand, line 86: Rather than having a really big comment, it would be better to wrap it in
```
if log_enabled!(Debug) {
...
}
```

src/docker/v2/handlers/blobs.rs
```diff
@@ -96,21 +97,45 @@ pub async fn handle_get_blobs(
 
     debug!("Reading blob contents..");
 
-    let mut blob_content_buf = Vec::new();
-    let blob_content_len = match art_reader.read_to_end(&mut blob_content_buf) {
+    let mut blob_content = Vec::new();
```
mgrand, line 100: Did you run clippy?  You should. Clippy usually asks you to substitute `vec![]` for this

mgrand (approved):

PR #200, opened 2022-01-07
Use new website repo name

prince-chrismc:


.github/workflows/rust.yml
```diff
@@ -84,8 +84,8 @@ jobs:
             mkdir -p pyrsia/usr/local/bin
             cp target/debug/pyrsia_node pyrsia/usr/local/bin
             cp target/debug/pyrsia pyrsia/usr/local/bin
-            chmod 755 pyrsia/usr/local/bin/pyrsia_node
-            chmod 755 pyrsia/usr/local/bin/pyrsia
+            chmod 777 pyrsia/usr/local/bin/pyrsia_node
+            chmod 777 pyrsia/usr/local/bin/pyrsia
```
mgrand, line 88: I don't know the context of this, but usually making anything writable be anyone is a risky thing to do. Perhaps you should consider an alternative?

mgrand (approved):

.github/workflows/rust.yml
sbtaylor15: 755 was working fine in the demo.yml workflow.  We should keep 755.

prince-chrismc: Sorry I did this on the side trying to fix https://github.com/pyrsia/pyrsia/runs/4742878764?check_suite_focus=true#step:7:51 but I am not sure it does

Excellent point, I was just trying to debug https://github.com/pyrsia/pyrsia/runs/4742813554?check_suite_focus=true#step:3:4

Seems to be two more bugs 😭

prince-chrismc:
The problem was deeper than I expected marking this as draft

sbtaylor15:
@prince-chrismc what is needed to wrap up this PR?

prince-chrismc:
I havent looked into it. Main failed with the PR being merged so not everything is hunky dory 🤷

prince-chrismc:
Mains magically passing 🤷

PR #220, opened 2022-01-13
Docker manifest to package version 2

mgrand:
This completes the logic for extracting package version metadata from docker manifests by handling manifests that conform to schema version 2.


sbtaylor15 (approved):
No changes needed

PR #222, opened 2022-01-13
Install openssl using brew

### LINKED ISSUE #221: build (macos-latest) is failing due to missing openssl library

sbtaylor15:
Need to use openssl from brew instead of the libreSSL which is the OS library.  Rust expects the openssl functions and not libreSSL ones.  Resolves #221


mgrand:
@sbtaylor15 My PR cannot be merged until after yours is. You will need to change the word "dummy" to "placeholder".

efrisch (approved):
👍

PR #226, opened 2022-01-14
Project version stored in document store

mgrand:
Previously we were just logging the PackageVersion metadata extracted from docker manifests.

Now call the metadata manager to store the metadata in the document store.


mgrand:
I discovered that there is no good way to start unit tests with an empty database in a dev environment. Rather than create a solution to that problem, I worked around it by adding a feature that was not needed yet. When the metadata manager tries to add a piece of metadata to the document store and the document store complains that it is a duplicate, the metadata manager treats that as a normal outcome and returns the contents of the duplicated metadata.

src/docker/v2/handlers/manifests.rs
```diff
@@ -687,7 +743,7 @@ mod tests {
     #[test]
     fn happy_put_manifest() -> Result<(), Box<dyn StdError>> {
```
betarelease, line 744: This name does not indicate what the happy path is can you please rename it?

Suggestions:
test_saves_manifest_with_(specified properties)
test_put_manifest_with_package_version


happy, sad paths are very broad names and do not communicate to me what is actually expected as input and what is the expected result. I look at test names to explain that to me without having to read the entire test.

src/node_manager/handlers.rs
```diff
@@ -131,11 +164,15 @@ mod tests {
...
-        //cleaning up
-        fs::remove_dir_all(ART_MGR_DIR)
-            .context(format!("Error cleaning up directory {}", path.clone()))?;
-
         println!("put_and_get_artifact_test ended !!");
         Ok(())
     }
+
+    #[test]
+    fn metadata_test() {
```
betarelease, line 172: Another instance where the test name does not describe to me, what is actually happening and expected.

I prefer this to be named as

test_public_key_pair_is_empty
test_private_key_pair_is_empty_(as a result of adding whatever metadata was added)

src/docker/v2/handlers/manifests.rs
```diff
@@ -108,24 +113,77 @@ pub async fn handle_put_manifest(
                 "Created PackageVersion from manifest: {:?}",
                 package_version
             );
+            if let Err(err) = sign_and_save_package_version(&mut package_version) {
```
betarelease, line 116: I did not see a test change for this error handling. Is there a way to test this? If this is not testable in the way it is written can we change it so that it becomes testable and add a corresponding test.

src/metadata_manager/metadata.rs
```diff
@@ -15,160 +15,367 @@
...
-    /// Get an iterator over the packages associated with the namespace identified by the given
-    /// namespace ID.
-    fn get_packages_by_namespace_id(
+pub struct Metadata {
+    //TODO Add a trust manager to decide if metadata is trust-worthy
+    package_type_docs: DocumentStore,
+    package_version_docs: DocumentStore,
+    untrusted_key_pair: SignatureKeyPair,
+}
+
```
fishseabowl, line 125: what is the different of package_type_docs and package_version_docs? Thanks

src/metadata_manager/metadata.rs
```diff
@@ -15,160 +15,367 @@
...
+            .license_text_mimetype(license_text_mimetype)
+            .license_url(license_url)
+            .metadata(pv_metadata)
+            .creation_time(creation_time)
+            .modified_time(modified_time)
+            .tags(tags)
+            .description(description)
+            .artifacts(artifacts)
+            .build()?;
+        let algorithm = signed::JwsSignatureAlgorithms::RS384;
```
fishseabowl, line 357: RS384 or RS512?

src/node_manager/handlers.rs
mgrand: @betarelease I renamed it to `test_that_a_metadata_manager_is_created_and_accessible`

src/docker/v2/handlers/manifests.rs
mgrand: @betarelease For this to fail, we would need to have a disk full or some other kind of I/O error. None of our code is currently being tested for handling this type of external problem. That is really outside the usual scope for unit tests.

Do you want us to start using mocking tools emulate this sort of thing? I am not sure what tools are available to emulate I/O errors and other problems that come from outside the program.

src/metadata_manager/metadata.rs
mgrand: @fishseabowl PackageVersion and PackageType are different types of metadata that are defined in the architecture document. 

A package type is a type of package that we will support such as docker, conan, npm, ...

A package (not yet implemented) describes a binary that is released with version numbers. A docker image named ubuntu or alpine would be described by a package.

A package version describes a specific release of a package. For example ubuntu:v20.04 would be described by a package version

src/metadata_manager/metadata.rs
mgrand: @fishseabowl For the purposes of these tests, it should not matter which signature algorithm is being used. I thought it would be best to use different algorithms so as to exercise different execution paths.

fishseabowl (approved):
Can we change untrusted_key_pair to another name? untrusted looks not good.

mgrand:
@fishseabowl 
> Can we change untrusted_key_pair to another name? untrusted looks not good.

It is exactly what it says it is. A key pair that is not trusted. It is intended only for testing. Right now we do not yet have trusted keys based on identities stored in the blockchain. 

We do not yet have trusted key pairs. We need to be able to test with signed metadata. For now we can unit test with a key pair that will not be trusted. This key is not really useful for any purpose other than unit testing things.  

Once we have trusted key pairs there will be additional tests. We may even remove the untrusted key pair. For now, we need it.

src/docker/v2/handlers/manifests.rs
```diff
@@ -108,24 +113,77 @@ pub async fn handle_put_manifest(
...
+fn put_manifest_response(
+    name: String,
+    hash: String,
+) -> Result<warp::http::Response<&'static str>, Rejection> {
+    Ok(
+        match warp::http::response::Builder::new()
+            .header(
+                LOCATION,
+                format!(
+                    "http://localhost:7878/v2/{}/manifests/sha256:{}",
```
efrisch, line 140: Long term, I think this needs to be a configurable uri.

src/document_store/document_store.rs
```diff
@@ -65,24 +65,107 @@ pub struct IndexSpec {
...
+            "At least one index specification is required when creating a DocumentStore."
+                .to_string(),
+        ))
+    } else {
+        Ok(())
+    }
+}
+
+fn collection_name_to_file_name(name: &str) -> String {
+    let mut s = name.to_string();
```
efrisch, line 139: `format!("{}.db", name)`

src/document_store/document_store.rs
```diff
@@ -65,24 +65,107 @@ pub struct IndexSpec {
...
+    document_store: &DocumentStore,
+) -> anyhow::Result<Catalog, DocumentStoreError> {
+    let raw_key = serialized_catalog_key()?;
+    let raw_doc_store = document_store
+        .unqlite
+        .kv_fetch(raw_key)
+        .map_err(DocumentStoreError::UnQLite)?;
+    let catalog: Catalog = bincode::deserialize(&raw_doc_store)?;
+    if catalog != document_store.catalog {
+        warn!("Stored catalog fof document store collection {} is different than expected. This may cause future errors.", catalog.name)
```
efrisch, line 154: Typo? fof == for?

src/document_store/document_store.rs
```diff
@@ -777,4 +786,9 @@ mod tests {
             .expect("Should have found a document."); // expect Some
         assert_eq!(doc.to_string(), res);
     }
+
+    fn append_random(name: &str) -> String {
+        let mut rng = rand::thread_rng();
+        format!("{}{}", name, rng.next_u32())
```
efrisch, line 792: Yeah. Like that.

efrisch (approved):
👍

Issue #209, opened 2022-01-11
Resolve clippy warnings.


mgrand:
We are getting many warnings from clippy about files that no one is actively working on. This makes it difficult to see the warnings about things that are actively being worked on.
This task is to go through the warnings and resolve them.

PR #231, opened 2022-01-17
fix clippy warnings #209

erwin1:
Fixed clippy warnings by
- running cargo fix
- fixing a number of warnings manually
- ignoring some warnings that I think we can live with (the module inceptions)

I also made the clippy github action check fail in case of warnings so we don't build up so many warnings again.

### DIFF
--- a/src/docker/v2/handlers/manifests.rs
+++ b/src/docker/v2/handlers/manifests.rs
@@ -320,7 +320,6 @@ const MEDIA_TYPE_BLOB_GZIPPED: &str = "application/vnd.docker.image.rootfs.diff.
 const MEDIA_TYPE_SCHEMA_1: &str = "application/vnd.docker.distribution.manifest.v1+json";
 const MEDIA_TYPE_IMAGE_MANIFEST: &str = "application/vnd.docker.distribution.manifest.v2+json";
 const MEDIA_TYPE_MANIFEST_LIST: &str = "application/vnd.docker.distribution.manifest.list.v2+json";
-const MEDIA_TYPE_CONFIG_JSON: &str = "application/vnd.docker.container.image.v1+json";
 
 fn package_version_from_schema1(
     json_object: &Map<String, Value>,
@@ -804,6 +803,8 @@ mod tests {
     }
 
     #[test]
+    const MEDIA_TYPE_CONFIG_JSON: &str = "application/vnd.docker.container.image.v1+json";
+
     fn package_version_from_image_manifest() -> Result<(), anyhow::Error> {
         let json_bytes = Bytes::from(MANIFEST_V2_IMAGE);
         let hash: Vec<u8> = raw_sha512(json_bytes.to_vec()).to_vec();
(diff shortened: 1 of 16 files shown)


src/docker/v2/handlers/manifests.rs
```diff
@@ -584,10 +583,9 @@ fn invalid_manifest<T>(json_string: &str) -> Result<T, anyhow::Error> {
 mod tests {
     use super::*;
     use crate::node_manager::handlers::ART_MGR;
-    use bytes::Bytes;
     use futures::executor;
     use serde::de::StdError;
-
+    const MEDIA_TYPE_CONFIG_JSON: &str = "application/vnd.docker.container.image.v1+json";
```
mgrand, line 588: This is identical to the line you deleted from 323. It looks like it belongs back there.  It there is a reason to put it hear, then ignore this comment.

src/document_store.rs
```diff
@@ -14,4 +14,5 @@
    limitations under the License.
 */
 
+#[allow(clippy::module_inception)]
```
mgrand, line 17: Please delete this annotation. This will be fixed with a real use in my next PR.

src/node_manager/model/cli.rs
```diff
@@ -26,8 +26,8 @@ pub struct Status {
 
 impl std::fmt::Display for Status {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        writeln!(f, "Connected Peers Count:   {}", self.peers_count);
-        writeln!(f, "Artifacts Count:         {}", self.artifact_count);
+        if let Err(_e) = writeln!(f, "Connected Peers Count:   {}", self.peers_count) {}
+        if let Err(_e) = writeln!(f, "Artifacts Count:         {}", self.artifact_count) {}
```
mgrand, line 30: The writeln!s are not conditional in the original version. This looks like a change to the logic, not just a clippy fix.

mgrand (approved):

src/node_manager/model/cli.rs
erwin1: The warning was this:
```
warning: unused `std::result::Result` that must be used
```

We were implicitly ignoring the Err variant of the result, which triggered the warning. So I decided to explicitly handle and ignore the error.

Maybe even better would be to compose the full string and write it using 1 write! macro, so the error can be properly returned as well? (which it already does now in the last of the 3 write's)

src/docker/v2/handlers/manifests.rs
erwin1: It was moved because it was only used in tests.

src/document_store.rs
erwin1: Ok, although you might just remove it in that PR then (or the other way round depending of what gets merged first)

I enabled failure on warnings, so this PR cannot be merged if I remove this.

src/node_manager/model/cli.rs
tiainen: The error of the last `writeln` is indeed propagated back to the caller. The simplest way to have the same behavior on the first two statements is to use the `?` operator.
```suggestion
        writeln!(f, "Artifacts Count:         {}", self.artifact_count)?;
```

tiainen (dismissed):

sbtaylor15 (approved):
/lgtm

tiainen (approved):

PR #237, opened 2022-01-19
Refactor Header, Block and Blockchain data struct and related methods

fishseabowl:
1. Refactor Header structure to support lightweight nodes quickly verify the validity of the blockchain.
2. All hashes in blockchain are temporarily encoded as a MultiHash, a self-describing hash format.
    Pros: Flexibility, Interoperability, Security.
    Cons: Overhead, Complexity, Lack of default
    Ref:
    https://ethresear.ch/t/multihashing-in-ethereum-2-0/4745
    https://datatracker.ietf.org/doc/html/draft-multiformats-multihash
3. Add unit test


src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
+            ),
         }
     }
-}
 
-pub struct BlockV1 {
-    pub header: Header,
-    pub transactions: Vec<Transaction>,
+    pub fn verify(&self) -> bool {
+        let pubkey = match libp2p::identity::ed25519::PublicKey::decode(&self.signature.pubkey) {
```
chb0github, line 77: Plucking data from the data stream should be done externally to the block chain. The block chain should have no knowledge or care of how it got data

chb0github: Alternatively to `match` you can do
`and_then(|v| => v)` and there is also a mapper for error

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
+            transaction_hash: hash,
+            signature: Signature::new(&bincode::serialize(&hash).unwrap(), ed25519_keypair),
+        }
+    }
 }
 
-pub fn signature(keypair: &identity::ed25519::Keypair, msg: &[u8]) -> Vec<u8> {
-    (*keypair).sign(msg)
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct PartialTransaction {
```
chb0github, line 122: What's a partial transaction?

fishseabowl: Hash and signature require other fields to be ready.

src/block_chain/block.rs
fishseabowl: > Plucking data from the data stream should be done externally to the block chain

I agree with you. It just verifies that the signature is valid.

pyrsia_node/src/main.rs
```diff
@@ -23,7 +23,7 @@ extern crate pyrsia;
 extern crate tokio;
 extern crate warp;
 
-use pyrsia::block_chain::block_chain::Ledger;
+//use pyrsia::block_chain::block_chain::Ledger;
```
mgrand, line 26: We should not be commenting out code. If you need to see code from the past, use git. This applies to other code you have commented out.

src/block_chain/block_chain.rs
```diff
@@ -13,203 +13,101 @@
...
-            })
-            .map(|(nonce, hash, last_block)| Block {
-                id: last_block.id + 1,
-                hash,
-                previous_hash: last_block.hash.clone(),
-                timestamp: now,
-                data,
-                nonce,
-            })
+    pub fn genesis(&mut self, keypair: &identity::ed25519::Keypair) {
```
prince-chrismc, line 33: This could be in the `new` to keep things clean

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
 use std::time::{SystemTime, UNIX_EPOCH};
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum TransactionType {
+    Create,
+}
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Signature {
+    signature: Vec<u8>,
+    pubkey: [u8; 32], //identity::ed25519::PublicKey a byte array in compressed form
+}
```
mgrand, line 29: Perhaps you should also include the signature algorithm?

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
-    pub id: u64,
-    pub hash: String,
-    pub previous_hash: String,
-    pub timestamp: u128,
-    pub data: String,
-    pub nonce: u64,
+type TransactionSignature = Signature;
+type BlockSignature = Signature;
+
+pub fn sign(msg: &[u8], keypair: &identity::ed25519::Keypair) -> Vec<u8> {
```
mgrand, line 43: I don't know if it is a good idea to use libp2p as the source of our signing algorithms. Though it is likely we will continue to use it, we are not yet certain that we will continue to use it for MVP. I would suggest using openssl instead so that our choice of crypto library is not tied to our choice of networking library.

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
+    }
 }
-
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Transaction {
-    pub nonce: U256,
     pub trans_type: TransactionType,
     pub submmitter: Address,
-    pub signature: TransactionSignature,
+    pub timestamp: u64,
```
mgrand, line 96: This is **NOT** a **PORTABLE** data format!!! Data portability is important since we will have nodes running on different machines and also will share blocks will all Pyrsia nodes.

The source code is portable, but the data will not be. The [rust definition of numeric types](https://doc.rust-lang.org/stable/reference/types/numeric.html) makes no mention of whether it will use big-endian or little-endian byte order. That is the reason that Rust provides methods such as [u32::to_ne_bytes](https://doc.rust-lang.org/std/primitive.u32.html#method.to_ne_bytes) to access the bytes in their native order.

The simplest way to keep the numbers portable is to store them as strings. If you must use a binary format, you will have to choose big-endian or little-endian and store the numbers as an array of bytes.

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
-
-pub fn generate_ed25519() -> identity::ed25519::Keypair {
-    //RFC8032
-    identity::ed25519::Keypair::generate()
+impl Transaction {
+    pub fn new(
+        partial_transaction: PartialTransaction,
+        ed25519_keypair: &identity::ed25519::Keypair,
+    ) -> Self {
+        let hash = hash(&(bincode::serialize(&partial_transaction).unwrap()));
```
mgrand, line 108: Please do not use unwrap. If the serialize method returns an error, unwrap will panic. We should not panic outside of our startup code.

I suggest that instead you have new return Result<Self, Error> so that serialization errors can be handled more robustly.

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
+    pub timestamp: u64,
+    pub payload: Vec<u8>,
+    pub nonce: u128,
 }
 
-pub fn get_publickey_from_keypair(
-    keypair: &identity::ed25519::Keypair,
-) -> identity::ed25519::PublicKey {
-    (*keypair).public()
+impl PartialTransaction {
```
mgrand, line 130: Why do you need a `PartialTransaction` type?  Wouldn't it be simpler to compute the hash and signature inside of `Transaction::new`?

src/block_chain/block.rs
```diff
@@ -13,119 +13,147 @@
...
+            nonce,
+        }
+    }
 }
 
 pub fn verify(pubkey: &identity::ed25519::PublicKey, msg: &[u8], sig: &[u8]) -> bool {
     (*pubkey).verify(msg, sig)
 }
+
 impl Display for Block {
```
mgrand, line 154: If you want to create the illusion that we are storing the block as JSON, then why not store the block as JSON? If the issue is signing a JSON string, all of the mechanism you need to do that is in `src/signed/src/signed.rs`.

src/block_chain/block_chain.rs
```diff
@@ -13,203 +13,101 @@
...
+        let local_id = hash(&get_publickey_from_keypair(keypair).encode());
+        let genesis_block_header = Header::new(PartialHeader::new(
+            hash(b""),
+            local_id,
+            hash(b""),
+            0,
+            rand::thread_rng().gen::<u128>(),
+        ));
+        let genesis_block = Block::new(genesis_block_header, vec![], keypair);
+        self.blocks.push(genesis_block);
```
mgrand, line 43: Should there be something that explicitly identifies the block as a Genisis block?

src/block_chain/block_chain.rs
fishseabowl: Thanks for your comment. I think this method might be upgraded later to support different configurations. I prefer to keep it.

src/block_chain/block_chain.rs
```diff
@@ -13,203 +13,101 @@
...
+    identity::ed25519::Keypair::generate()
 }
 
-impl Display for BlockChain {
+impl Display for Blockchain {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         let json = serde_json::to_string_pretty(&self).expect("json format error");
         write!(f, "{}", json)
     }
 }
```
mgrand, line 83: The size of the string produced could be a problem in the future. Please look into ways to limit the size, perhaps by including only a maximum number of blocks.

src/block_chain/header.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+pub fn hash(msg: &[u8]) -> HashDigest {
+    HashDigest {
+        multihash: Code::Keccak256.digest(msg),
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct PartialHeader {
```
mgrand, line 60: I don't think you need a separate PartialHeader struct.

src/block_chain/block_chain.rs
prince-chrismc: That sounds reasonable

src/block_chain/block.rs
chb0github: Definitely a handy bit of meta-data

src/block_chain/block.rs
chb0github: I would abstract the whole thing as a trait on block and idependently implement it. You want it to be a totally isolated function.

I don't know if you can do something like this in rust, but what you want is a set of functions implemented on enums, and then you pass the enum representing the function. 

Basically, you want functions as enums

src/block_chain/block.rs
chb0github: He's absolutely right. BE vs LE is a forgotten subject and can't be overlooked. 
However, if it get's transferred over-the-wire in a format like json, then the number does indeed get transmitted as a string. 

@mgrand - portability will be a function of the encoding layer. The block and block chain should never concern itself with how data comes in-and-out of the app

src/block_chain/block.rs
chb0github: The real question is: When does the team move out of prototype mode and into production grade mode?

src/block_chain/block.rs
chb0github: I was asking myself the same question

src/block_chain/block.rs
mgrand: @chb0github Having had to do something like this, what you do is implement methods for an enum and then have the methods use a match expression to dispatch to logic specific to the enum variant. The compiler will tell you if any of the match expressions do not cover all the variants. Here is an example:
```
pub enum JwsSignatureAlgorithms {
    RS512,
    RS384,
}

impl JwsSignatureAlgorithms {
    /// Return a string that contains the JWS name of this.
    pub fn to_jws_name(&self) -> String {
        String::from(match self {
            JwsSignatureAlgorithms::RS512 => "RS512",
            JwsSignatureAlgorithms::RS384 => "RS384",
        })
    }

    /// Return a MessageDigest appropriate for the algorithm.
    fn as_message_digest(&self) -> MessageDigest {
        match self {
            JwsSignatureAlgorithms::RS512 => MessageDigest::sha512(),
            JwsSignatureAlgorithms::RS384 => MessageDigest::sha384(),
        }
    }
...
}
```

src/block_chain/block_chain.rs
chb0github: You know, maybe blocks should be Enums ? 

```rust
pub enum Block {
   Genesis,
   Other(data)
}
```
I kind of like the idea as it makes things clearer and easier to work with.

src/block_chain/block_chain.rs
chb0github: This seems to me for debug - to be used sparingly (or, dumped to disk)

src/block_chain/block_chain.rs
mgrand: Making entire blocks be enums gives us some messy choices about how to discriminate between the different block types in a way that can be included in the hash. I think it would be simpler to use non-parametric enum variants as the value of a block_type field.

src/block_chain/block.rs
chb0github: That's a way. I am thinking more like this:
```java
enum Action {
    DODGE {
        public boolean execute(Player player) {
            return player.isAttacking();
        }
    },
    ATTACK {
        public boolean execute(Player player) {
            return player.hasWeapon();
        }
    },
    JUMP {
        public boolean execute(Player player) {
            return player.getCoordinates().equals(new Coordinates(0, 0));
        }
    };

    public abstract boolean execute(Player player);
}
```

Match statements are a brute-force method. I am wondering if something more like the above is possible

mgrand: I would have very much preferred to write code like this, matching logic with variants in the declaration rather than at run time with the dispatch being specified in each method. That is not the way rust works. It would be a good addition to the language.

pyrsia_node/src/main.rs
fishseabowl: Thanks. I fix it.

src/block_chain/block.rs
fishseabowl: Sure. I think I'll create a new structure for the genesis block, including all metadata like signature algorithm, random number algorithm, key length...

src/block_chain/block.rs
fishseabowl: Sure. I agree with you too. There are still some areas that need to be improved and discussed, including signature algorithms, keys and key storage. For example, Bitcion's private key should ideally only be used once. When used up, the private key is worthless.

src/block_chain/block.rs
fishseabowl: Of course. I would add custom byte array.

src/block_chain/block.rs
fishseabowl: Hash and signature require other fields to be ready.

src/block_chain/header.rs
fishseabowl: @mgrand Hash and signature require other fields to be ready.

src/block_chain/block.rs
fishseabowl: I think it's a decentralized system. So storage and computation are very expensive. JSON is human-readable, but not computer-efficient. Blockchain would provide JSON API.

src/block_chain/block_chain.rs
fishseabowl: Of course. I think I'll create a new structure for the genesis block, including all metadata like signature algorithm, random number algorithm, key length...

src/block_chain/block.rs
mgrand: I think you are worrying about optimization too soon. Whatever performance challenges we have will not be caused by JSON. Both BigchainDB and Hyperledger Fabric use JSON and that has not been an issue. BigchainDB uses MongoDB for storing blocks. Hyperledger Fabric uses CouchDB. Both are JSON based document stores.

First let us have something that works, then we can worry about performance. I believe that the big things we will be concerned about will be consensus and propagation of blocks. JSON operations occur entirely within the computer, hundreds or thousands of times faster than persistent storage and perhaps 10,000 times faster than network operations. Based on the experience people have had with Fabric and BigchainDB, I think it is more important to keep things simple at this point than to try optimize something that may not make a noticeable difference.

src/block_chain/header.rs
mgrand: So make the "temporary" struct the value of a field in the complete struct, replacing the existing data fields. Then it is no longer temporary and it becomes clearer in the code what the hash and signature are based on.

src/block_chain/block_chain.rs
fishseabowl: Yes, this will be upgraded.

src/block_chain/block.rs
fishseabowl: I don't think we are talking about the same thing. Hyperledger Fabric serialize structured data using ProtoBuf.  Ref: https://github.com/hyperledger/fabric-protos-go/blob/main/common/common.pb.go. In Hyperledger Fabric,  CouchDB works as the Peer State Database. https://hyperledger-fabric.readthedocs.io/en/release-2.2/couchdb_as_state_database.html

prince-chrismc (approved):
This looks like a great step!

The new structures are much more complete and they replace the old ones as we discussed in #192.

Can wait to see the implementation in main get updated next 🚀

mgrand (approved):

PR #246, opened 2022-01-20
Artifact manager moves files from directories

mgrand:
This implements an artifact manager feature that will be used to move files from the staging directory used by synapse to the artifact manager.

There is also a change to encode hashes as file names using base58 rather than hex.

Functions that were in the ArtifactManager struct that should not have been were removed.


betarelease:
As I mentioned in my comments on the issue can you please separate this into two different PRs - one for changing to use bs58 and another that covers the move functionality. As it is we should not mix the two concerns in the same PR.

src/artifact_manager.rs
```diff
@@ -453,6 +382,132 @@ impl<'a> ArtifactManager {
     }
 }
 
+fn compute_hash_of_file<'a>(
```
betarelease, line 385: Please add tests for each of these new methods before merging this PR. This is a lot of code without any tests - I see a lot of magic numbers and also lot of assumptions which I would like to see tested in detail.

mgrand:
Closing this without merging as I have been requested to separate it into two different pull requests

src/artifact_manager.rs
```diff
@@ -175,17 +176,15 @@ impl<'a> Hash<'a> {
     }
 
     fn encode_bytes_as_file_name(bytes: &[u8]) -> String {
-        hex::encode(bytes)
+        bs58::encode(bytes).into_string()
```
betarelease, line 179: I see that this is a one line change but the repercussions could be wider.

Please cover this with tests that explain whey we cannot use hex any longer and need to move to this bs58 crate.


Please do not merge this PR without such tests being included in the PR.

mgrand: @betarelease This is covered by existing tests. The existing tests that exercise our ability to push and pull artifacts use this to translate hash values to file names.

PR #247, opened 2022-01-21
Artifact manager moves files from directories

mgrand:
This PR adds a move_from method to the artifact manager we can use to move files from Synapse's staging directory into the artifact manager.

There is also some clean-up. Functions that were declared as part of the ArtifactManager struct impl but shouldn't have been are moved out. 

There is also a test in handlers that has been modified. It was looking at way the artifact manager was organizing files, which is an internal implementation detail that will change. I modified the test to not rely on the artifact manager's internals.


src/artifact_manager.rs
```diff
@@ -273,41 +273,27 @@ impl<'a> ArtifactManager {
...
+            .filter_entry(is_not_hidden)
             .filter_map(|file| file.ok())
         {
-            if file.metadata().unwrap().is_file() {
-                total_files += 1;
+            if let Ok(metadata) = file.metadata() {
+                if metadata.is_file() {
+                    total_files += 1;
+                }
             }
```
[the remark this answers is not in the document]
mgrand: Yes. We are not currently caching any counts. It is likely that artifacts will be added and deleted asynchronously. 

To scale, we may want to cache approximate counts in the future, but is it too early to be optimizing something like that.

src/artifact_manager.rs
```diff
@@ -273,41 +273,27 @@ impl<'a> ArtifactManager {
...
         for file in WalkDir::new(repository_path)
             .into_iter()
-            .filter_entry(|path| Self::is_not_hidden(path))
+            .filter_entry(is_not_hidden)
             .filter_map(|file| file.ok())
         {
-            if file.metadata().unwrap().is_file() {
-                total_files += 1;
+            if let Ok(metadata) = file.metadata() {
+                if metadata.is_file() {
```
MitaliBo, line 289: so are only counting metadata files?

src/node_manager/handlers.rs
```diff
@@ -131,24 +130,12 @@ mod tests {
...
-        push_art_path.push("SHA256");
-        push_art_path.push(hex::encode(GOOD_ART_HASH));
-        push_art_path.set_extension("file");
-        println!("reading artifact path is: {}", push_art_path.display());
-        let content_vec = fs::read(push_art_path.as_path()).context("reading pushed file")?;
-
-        let test_art_path = PathBuf::from(path.as_str());
-        let actual_content_vec = fs::read(test_art_path.as_path()).context("reading test file")?;
-
-        assert_eq!(content_vec.as_slice(), actual_content_vec.as_slice());
```
MitaliBo, line 149: why this validation was removed from test?

src/artifact_manager.rs
mgrand: No. All we are counting all artifacts. This is an unfortunate case of two unrelated things having the same name of metadata. 

In the rust std::fs crate, the way that you learn if a file exists or what kind of file it is, is to call the File struct's metadata method to get a `Result<Metadata>`. See https://doc.rust-lang.org/std/fs/struct.Metadata.html.

Let me know if you have any ideas on how to make this less confusing.

src/node_manager/handlers.rs
mgrand: It was looking at the internal details of the artifact manager. The test should not care about how the artifact manager stores artifacts. It should just care that it can push an artifact to the artifact manager and then pull the same artifact from the artifact manager.

The current organization is not scalable and will need to be changed for MVP. This test in the handlers.rs should not break if we change the implementation details of the artifact manager.

src/artifact_manager.rs
MitaliBo: Metadata is the only way unfortunately

MitaliBo (approved):

src/artifact_manager.rs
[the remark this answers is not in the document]
fishseabowl: I am confused about the artifact manager synchronization mechanism. How to synchronize artifacts? Is add or delete an atomic operation? If you have documentation or code on this, please share with me. I'm very grateful.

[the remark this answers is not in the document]
mgrand: @fishseabowl There is no synchronization mechanism. It is possible for multiple artifacts to be pushed at the same time. If the future, when we are enforcing disk usage quotas, it will also be possible for artifacts to be evicted at the same time, though artifacts currently being pulled may be exempt from eviction.

What synchronization mechanism are you expecting?

src/node_manager/handlers.rs
MitaliBo: it is just looking into matching the contents and that is the way to validate if that is the same artifact, i think test in artifact_manager itself doing the same

src/artifact_manager.rs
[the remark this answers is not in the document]
fishseabowl: @mgrand As you mentioned "It is likely that artifacts will be added and deleted asynchronously." So I'm asking about synchronization mechanisms and atomic operations.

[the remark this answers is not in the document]
mgrand: The only atomic operation involved is currently the rename of files, which we rely on the native file system to provide atomically. While a file is being pushed into the artifact manager, it has a temporary name that prevents it from being mistaken for an available artifact. When the artifact is fully copied into the file, the file is renamed. The rename makes it available.

src/artifact_manager.rs
```diff
@@ -328,61 +314,49 @@ impl<'a> ArtifactManager {
...
+            bail!(
+                "{} does not have the expected hash value {}:{}\nActual hash is {}:{}",
+                path.display(),
+                expected_hash.algorithm,
+                hex::encode(computed_hash),
+                expected_hash.algorithm,
+                hex::encode(expected_hash.bytes)
+            )
+        }
+        let target_path = self.file_path_for_new_artifact(expected_hash);
```
fishseabowl, line 358: self.file_path_for_new_artifact(expected_hash) or file_path_for_new_artifact(expected_hash)

src/artifact_manager.rs
```diff
@@ -328,61 +314,49 @@ impl<'a> ArtifactManager {
         // Write to a temporary name that won't be mistaken for a valid file. If the hash checks out, rename it to the base name; otherwise delete it.
         let tmp_path = tmp_path_from_base(&base_path);
 
-        match Self::create_artifact_file(&tmp_path) {
+        match create_artifact_file(&tmp_path) {
             Err(error) => file_creation_error(&tmp_path, error),
             Ok(out) => {
                 debug!("hash is {}", expected_hash);
-                let mut hash_buffer: [u8; 128] = [0; 128];
+                let mut hash_buffer = [0; HASH_BUFFER_SIZE];
```
fishseabowl, line 321: hash_buffer = [0; HASH_BUFFER_SIZE];  or hash_buffer = [0u8; HASH_BUFFER_SIZE];?

mgrand: @fishseabowl because of type inference from its being use as a parameter of do_push, it its type is [u8; HASH_BUFFER_SIZE].
There is a trade off here between conciseness of code and ease of reading outside of an IDE. The deleted version of the statement was written when I was new to Rust. When I created a symbolic name for the size of the buffer, I decided to leave out the explicit type declaration. I think it is more readable this way, but apparently you felt confused. Do you think I should put the explicit type declaration back in?

src/artifact_manager.rs
mgrand: @fishseabowl It is self.file_path_for_new_artifact. It is part of the impl because it has to get the path of the parent directory from the ArtifactManager struct.

efrisch (approved):
👍

src/artifact_manager.rs
```diff
@@ -453,6 +383,132 @@ impl<'a> ArtifactManager {
...
+    path: &Path,
+    out: File,
+    hash_buffer: &'b mut [u8; HASH_BUFFER_SIZE],
+) -> Result<&'b [u8], Error> {
+    let mut buf_writer: BufWriter<File> = BufWriter::new(out);
+    let mut digester = expected_hash.algorithm.digest_factory();
+    let mut writer = WriteHashDecorator::new(&mut buf_writer, &mut digester);
+
+    copy_from_reader_to_writer(reader, path, &mut writer)
+        .with_context(|| format!("Error writing contents of {}", expected_hash))?;
```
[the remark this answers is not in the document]
mgrand: Yes. Artifacts are identified by their hash.

fishseabowl (approved):
it looks fine to me. thanks

Issue #213, opened 2022-01-12
Enable the docker integration to use ProjectVersion structs from the metadata manager to retrieve manifests from the artifact manager.


mgrand:
Given that we are storing docker manifests in the artifact managers and also storing PackageVersion metadata that describes the manifests from a Pyrsia point of view (see #212), enhance the node to fetch the manifest from the artifact manager rather than dockerHub.

PR #248, opened 2022-01-21
Docker integration uses metadata

mgrand:
This PR is a first installment on issue #213.  It is not linked so that merging this will not close the issue.

All that this does is add support for namespace metadata. Namespaces will be needed for the docker integration to find the PackageVersion that will tell it what artifacts correspond to a named docker image and tag.


efrisch (approved):
👍

MitaliBo (approved):

PR #255, opened 2022-01-21
pyrsia CLI status error

### LINKED ISSUE #239: Pyrsia CLI get status fails
```
pyrsia config --add localhost:7878

pyrsia node -p
Connection Successfull !! {}

pyrsia node -s
Error: error decoding response body: missing field `peers_count` at line 1 column 49
```

MitaliBo:
port change to 7888
somehow status api was overridden by previous changes

Closes #239

### DIFF
--- a/pyrsia_node/src/main.rs
+++ b/pyrsia_node/src/main.rs
@@ -55,7 +55,7 @@ use tokio::{
 };
 use warp::Filter;
 
-const DEFAULT_PORT: &str = "7888";
+const DEFAULT_PORT: &str = "7878";
 
 #[tokio::main]
 async fn main() {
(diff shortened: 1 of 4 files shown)


pyrsia_node/src/main.rs
```diff
@@ -55,7 +55,7 @@ use tokio::{
 };
 use warp::Filter;
 
-const DEFAULT_PORT: &str = "7878";
+const DEFAULT_PORT: &str = "7888";
```
erwin1, line 58: if we change the default port, we need to change a number of other things (in fact, I'd prefer a separate issue and PR). Things I think about that needs to change as well:
- the Dockerfile 
- the README
- the message to configure Docker that the linux installers outputs

But there might be other places as well.

MitaliBo: Agreed created new issue for this #259

erwin1 (dismissed):

erwin1 (approved):

fishseabowl (approved):
If possible, please add unit tests. thanks

Issue #253, opened 2022-01-21
Fix dockerized version of Pyrsia node


erwin1:
2 little things are missing to make the dockerized version work:
- make the Pyrsia server host configurable so it can bind on 0.0.0.0 instead of 127.0.0.1
- install the latest ca-certificates in the container

erwin1:
@sbtaylor15 thanks. openssl isn't installed by the Dockerfile, but `libssl1.1` is. 

And `apt-get install ca-certificates` fixes the issue. Unless you think `openssl` should be installed instead?

sbtaylor15:
@erwin1 openssl is now a required dependency.  When we do `apt-get install pyrsia` openssl gets installed if missing.  Realistically, the dockerized version shouldn't do a cargo build but install pyrsia using apt-get.  That way we are testing what the end user (native Ubuntu) is doing.

sbtaylor15:
Added docker build in PR #260 to create image and push to dockerhub based on apt-get installer.

Issue #181, opened 2021-12-31
Investigate alternative to OpenSSL


mgrand:
The signature code currently uses the OpenSSL library to create and verify RSA signatures.  This causes use to have the OpenSSL library installed to build. We should investigate the availability of a pure Rust alternative that will not require an external library to be installed for builds to succeed.

fishseabowl:
1. The Linux Foundation has a shared cryptographic library project: Hyperledger Ursa.
Reference: https://www.hyperledger.org/use/ursa
Maybe we can try this.

2. The signature algorithm in the blockchain currently uses the libp2p ed25519 algorithm.
├── libp2p-core v0.31.0
│   ├── ed25519-dalek v1.0.1
│   │   ├── curve25519-dalek v3.2.0
│   │   ├── ed25519 v1.3.0

mgrand:
Since getting this done will facilitate integration testing with Windows based laptops, I am moving this to the to-do column.

sbtaylor15:
Closing since this will be addressed in the future under #511

Issue #178, opened 2021-12-30
Build failed if openssl is not installed


fishseabowl:
I found if Openssl is not installed, it would be failed to build pyrsia node.  Could we add support/pre-installed packages on readme.md?

Look like:
First
$ rustup update
Once done, finish installing the support packages:
$ brew install openssl@1.1

sbtaylor15:
We can add this as a requirement in the .deb file so it will get installed as part of the `apt-get install pyrsia`

mgrand:
@fishseabowl What are we doing that requires OpenSSL? Is it something that is available using a Rust-based crate?

fishseabowl:
When I run cargo run on pyrsia node, I got the below information:
**error: failed to run custom build command for `openssl-sys v0.9.72`**
Caused by:
  process didn't exit successfully: `.../pyrsia/target/debug/build/openssl-sys-6398316d2915563b/build-script-main` (exit status: 101)
  --- stdout
  cargo:rustc-cfg=const_fn
  cargo:rerun-if-env-changed=X86_64_APPLE_DARWIN_OPENSSL_LIB_DIR
  X86_64_APPLE_DARWIN_OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
  OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=X86_64_APPLE_DARWIN_OPENSSL_INCLUDE_DIR
  X86_64_APPLE_DARWIN_OPENSSL_INCLUDE_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
  OPENSSL_INCLUDE_DIR unset
  cargo:rerun-if-env-changed=X86_64_APPLE_DARWIN_OPENSSL_DIR
  X86_64_APPLE_DARWIN_OPENSSL_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_DIR
  OPENSSL_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-apple-darwin
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_apple_darwin
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=OPENSSL_STATIC
[...]

PR #256, opened 2022-01-21
#253 fix running in docker

erwin1:
fix running in docker by allowing to set a host to bind to in Pyrsia node

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN <<EOT bash
     set -e
     apt-get update
     apt-get install -y \
-        libssl1.1
+        openssl
     rm -rf /var/lib/apt/lists/*
 EOT
 COPY --from=dbuild /out/pyrsia_node /usr/local/bin/
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,7 @@ RUN <<EOT bash
     set -e
     apt-get update
     apt-get install -y \
+        ca-certificates \
         openssl
     rm -rf /var/lib/apt/lists/*
 EOT
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,7 @@ services:
   pyrsia:
     image: pyrsia/node
     build:
+      context: .
       target: node
     command: ["-h", "0.0.0.0", "-p", "7878"]
     stdin_open: true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     image: pyrsia/node
     build:
       target: node
-    command: ["-p", "7878"]
+    command: ["-h", "0.0.0.0", "-p", "7878"]
     stdin_open: true
     ports:
       - "7878:7878"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       target: node
-    command: ["-h", "0.0.0.0", "-p", "7878"]
+    command: ["-H" , "0.0.0.0", "-p", "7878"]
     stdin_open: true
     ports:
       - "7878:7878"
--- a/pyrsia_node/src/main.rs
+++ b/pyrsia_node/src/main.rs
@@ -55,6 +55,7 @@ use tokio::{
 };
 use warp::Filter;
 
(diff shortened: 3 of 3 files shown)


sbtaylor15 (approved):
Looks good for now.  We should skip building from scratch and run `apt-get install pyrsia` down the road.

docker-compose.yml
```diff
@@ -2,8 +2,9 @@ services:
   pyrsia:
     image: pyrsia/node
     build:
+      context: .
       target: node
-    command: ["-p", "7878"]
+    command: ["-h", "0.0.0.0", "-p", "7878"]
```
tiainen, line 7: Shouldn't the host parameter be capital h, or is it case insensitive? 
```suggestion
    command: ["-H" , "0.0.0.0", "-p", "7878"]
```

pyrsia_node/src/main.rs
```diff
@@ -55,6 +55,7 @@ use tokio::{
 };
 use warp::Filter;
 
+const DEFAULT_HOST: &str = "127.0.0.1";
 const DEFAULT_PORT: &str = "7878";
```
sbtaylor15, line 59: Default port should be changed to 7888

Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN <<EOT bash
     set -e
     apt-get update
     apt-get install -y \
-        libssl1.1
+        openssl
```
tiainen, line 40: I managed to get the pyrsia_node up and running via docker compose. However, when trying to pull an image, it failed with the following error:

    pyrsia_1  |  DEBUG pyrsia::docker::error_util   > Rejection: Rejection([MethodNotAllowed, RegistryError { code: Unknown("error sending request for url (https://auth.docker.io/token?client_id=Pyrsia&service=registry.docker.io&scope=repository:library/alpine:pull): error trying to connect: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1915: (unable to get local issuer certificate)") }])

If I add `ca-certificates` to the list of apt packages to install, the TLS error is gone.

pyrsia_node/src/main.rs
erwin1: I'd like to address these changes in a separate PR fixing issue #259 since it involves more than just this change. (README, Dockerfile, config default, docs, ... have to be changed as well)

tiainen (approved):

Dockerfile
```diff
@@ -37,7 +37,8 @@ RUN <<EOT bash
     set -e
     apt-get update
     apt-get install -y \
-        libssl1.1
+        ca-certificates \
+        openssl
```
chris-crone, line 41: @tiainen Any reason to install the whole of openssl over just libssl and ca-certificates?

tiainen: There was some discussion about this on the issue: https://github.com/pyrsia/pyrsia/issues/253#issuecomment-1018734622

It looks like the reason for installing `openssl` comes from this issue: https://github.com/pyrsia/pyrsia/issues/178
A follow-up issue was created to investigate if it can be replaced: https://github.com/pyrsia/pyrsia/issues/181

Personally, I have not run into issues when using `libssl1.1` instead. Everything still seems to be working. Maybe @sbtaylor15 can shed some light into the actual reason of why `openssl` is a requirement.

mgrand: We are using OpenSSL because we have code that relies on the library. The only way I knew to get the library was to install OpenSSL. If we can install just the library, that would be better.

chris-crone: Thanks for the context @tiainen! I agree with @mgrand that if we can reduce our requirements/dependencies then we should. When I wrote the original Dockerfile, I wasn't sure exactly what was needed but found the binary started with just libssl installed.

sbtaylor15: Redhat/Fedora refer to the ssl libraries as openssl and Debian/Ubuntu as libssl.  Just a carry over from my Redhat development.  If libssl1.1 works then lets go with that.  Lets merge this and I will replace openssl with libssl1.1 and ca-certificates everywhere in a separate PR.

sbtaylor15 (approved):
Looks good to merge.  I will create another PR for using libssl1.1 & ca-certificates everywhere.

PR #272, opened 2022-01-25
Added disk usage cli status 

MitaliBo:
Added Disk usage percentage and disk space allocated to store artifacts.
Added check for available disk space
Fixed counting of nodes in status


sbtaylor15 (approved):
/lgtm

sbtaylor15 (approved):
Looks good

src/node_api/handlers/swarm.rs
```diff
@@ -67,11 +66,18 @@ pub async fn handle_get_status(
         }));
     }
 
+    let disk_space_result = disk_usage();
+    if disk_space_result.is_err() {
+        return Err(warp::reject::custom(RegistryError {
+            code: RegistryErrorCode::Unknown(disk_space_result.err().unwrap().to_string()),
+        }));
+    }
+
```
mgrand, line 75: This function is getting large enough that it should be broken up into smaller functions.

src/node_manager/handlers.rs
```diff
@@ -63,7 +67,7 @@ fn log_static_initialization_failure<T: UnwindSafe>(
     match panic_wrapper {
         Ok(normal) => normal,
         Err(partially_unwound_panic) => {
-            error!("Initialization of {} paniced!", label);
+            error!("Initialization of {} panicked!", label);
```
mgrand, line 70: I think this was my mistake. Thank you for fixing it.

src/node_manager/handlers.rs
```diff
@@ -92,20 +96,43 @@ pub fn put_artifact(
...
-    info!("get_pyrsia_status started");
-
     ART_MGR
         .artifacts_count(ART_MGR_DIR)
         .context("Error while getting artifacts count")
 }
 
+pub fn get_space_available() -> Result<u64, anyhow::Error> {
+    let disk_used_bytes =
+        get_size(ART_MGR_DIR).context("Error while calculating the size of artifact manager")?;
```
mgrand, line 112: How much space the artifact manager is using is a calculation that should be done inside the artifact manager. Other modules should not be making assumptions about how the artifact manager manages storage. For example, in the future we might decide that this total should include storage that is not yet being used, but is reserved for a file that is being downloaded by bittorrent. That is something that will be known inside the artifact manager, but not here.

mgrand (commented):
Please see my comments for a recommended change.

src/node_manager/handlers.rs
MitaliBo: moved to Artifact manager

src/node_api/handlers/swarm.rs
MitaliBo: this is just a method which collect all the parameters for CLI "status" , for each field there is already a function defined separately, i don't think it needs to be broken up further

src/artifact_manager.rs
```diff
@@ -294,6 +295,10 @@ impl<'a> ArtifactManager {
         Ok(total_files)
     }
 
+    pub fn space_used(&self, repository_path: &str) -> Result<u64, Error> {
+        get_size(repository_path).context("Error while calculating the size of artifact manager")
```
mgrand, line 299: You shouldn't be passing `repository_path` as a parameter. It is available as `self.repository_path`

Also you will need to add a unit test to exercise it.

src/node_manager/handlers.rs
```diff
@@ -31,7 +30,7 @@ use std::panic::UnwindSafe;
 use std::str;
 use std::{fs, panic};
 
-const ART_MGR_DIR: &str = "pyrsia";
+pub const ART_MGR_DIR: &str = "pyrsia";
```
mgrand, line 33: This does not need to be public.

MitaliBo: done

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -41,7 +41,7 @@ pub async fn handle_node_ping() {
     let result = ping().await;
     let _resp = match result {
         Ok(resp) => {
-            println!("Connection Successfull !! {}", resp)
+            println!("Connection Successfull !!")
```
efrisch, line 44: Should these `println!`(s) be replaced with logger calls?

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -41,7 +41,7 @@ pub async fn handle_node_ping() {
...
         }
         Err(error) => {
             println!("Error: {}", error);
```
efrisch, line 47: This seems more critical than the one on line 44.

efrisch (approved):
Looks good to me. I only question a few println! calls.

pyrsia_cli/src/cli/handlers.rs
MitaliBo: well all these are command line outputs so that is why println as logger will not output it

pyrsia_cli/src/cli/handlers.rs
MitaliBo: same needed for cli output

mgrand (approved):

mgrand:
@MitaliBo I am merging this for you.

PR #289, opened 2022-01-26
The Artifact Manager Creates a Torrent When an Artifact is Pushed to it.

mgrand:


src/artifact_manager.rs
```diff
@@ -249,16 +255,20 @@ impl<'a> Write for WriteHashDecorator<'a> {
...
-/// `68efadf3184f20557aa2bbf4432386eb79836902a1e5aea1ff077e323e6ccbb4.file`
+// `680fade3184f20557aa2bbf4432386eb79836902a1e5aea1ff077e323e6cab34`
 ///
 /// For now, all files will have the `.file` extension to signify that they are simple files whose
 /// contents are the artifact having the same hash as indicated by the file name. Other extensions
 /// may be used in the future to indicate that the file has a particular internal structure that
 /// Pyrsia needs to know about.
 pub struct ArtifactManager {
     pub repository_path: PathBuf,
+    peer_id: RwLock<Option<PeerId>>,
```
fishseabowl, line 266: I am confused about sync.RwLock peer_id . Do you think different nodes can read/write this value at the same time?

mgrand: @fishseabowl  No. That is not the issue. The problem stems from the fact that we will be using an instance of the ArtifactManager struct that is statically allocated in handlers.rs. The mechanism we are using to do the static initialization is the `lazy_static!` macro. To keep the integrity of Rust's borrowing rules, the macro only allows us to have an immutable reference to the ArtifactManager instance.

Since peer_id and Kademlia dht are created by the handlers module, it is also up to the handlers module to make these visible to the ArtifactManager by calling a set method. Normally set methods require a mutable reference to a struct, but that is not possible for ArtifactManager. 

This is not a unique situation. The rust documentation references a design pattern called "internal mutability" to describe needed to have a struct that modifies itself, even though there are no mutable references to it. I originally tried the using the Cell and RefCell structs to implement internal mutability. Because they are referenced by something that has the static lifetime, Rust complained that they can be accessed by multiple threads and that Dell and RefCell are not thread safe. It is true that our intended using pattern is thread safe (write once, ready many times), but Rust does not recognize that.

The thread safe solutions recommended for multi-threaded internal mutability of non-primitive values are Mutex and RwLock. Mutex only permits access to one thread at a time. We want to allow multiple concurrent reads. Mutex does not permit that. RwLock allows either one mutable borrow or multiple immutable borrows. RwLock fits this pattern, so that is what I used.

efrisch (approved):
👍

src/artifact_manager.rs
```diff
@@ -404,6 +449,43 @@ fn compute_hash_of_file<'a>(
...
+        }
+        Err(error) => bail!(
+            "Error writing torrent to file {}: {}",
+            torrent_path.display(),
+            error
+        ),
+    }
+}
+
+const DEFAULT_TORRENT_PIECE_SIZE: i64 = 1048576;
```
fishseabowl, line 474: how about u64? or this value might be negative?

mgrand: I would prefer to be using u64, but the torrent create that I am using has a piece size that is i64. This is the size of the pieces that the bittorrent will use to download a file, not the size of the file. Typical values for this are no larger than 2^24, so I do not see this as a problem. Do you have any more questions before approving?

fishseabowl (approved):
Thanks for the detailed explanation.

PR #295, opened 2022-01-28
Added Configuration information to Blockchain

fishseabowl:
1. Add configuration information to blockchain
2. define genesis block struct
3. Add unit test


pyrsia_node/src/main.rs
```diff
@@ -196,7 +198,7 @@ async fn main() {
     tokio::spawn(server);
     let tx4 = tx.clone();
 
-    let raw_chain = block_chain::Blockchain::new();
+    let raw_chain = block_chain::Blockchain::new(&local_key);
```
mgrand, line 201: You should not have functions over 100 lines long. Breaking this into sub-functions will make this more readable, easier for you to find bugs and if this is ever maintained by someone else, easier for them to understand. Especially if the sub-functions have meaningful names.

Most IDEs have a way to automate the refactoring of a chunk of code into a separate function.

src/block_chain/block.rs
```diff
@@ -38,7 +38,7 @@ impl Signature {
 }
 
 type TransactionSignature = Signature;
-type BlockSignature = Signature;
+pub type BlockSignature = Signature;
 
 pub fn sign(msg: &[u8], keypair: &identity::ed25519::Keypair) -> Vec<u8> {
```
mgrand, line 43: What is being signed here. It is not obvious. If it is the block, you will need to add a way to securely manage keys. If it is the transaction, that is a problem. Transactions should be signed before they get to the blockchain.

src/block_chain/block_chain.rs
```diff
@@ -13,91 +13,157 @@
...
 use libp2p::identity;
 use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::fmt::{Display, Formatter};
 
+/// BlockchainId identifies the current chain
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct Blockchain {
-    pub blocks: Vec<Block>,
+pub enum BlockchainId {
```
mgrand, line 26: Unclear what this is for.

src/block_chain/block_chain.rs
```diff
@@ -13,91 +13,157 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Blockchain {
+    pub genesis_block: GenesisBlock,
+    pub blocks: Vec<Block>,
+}
+
+impl Blockchain {
+    pub fn new(keypair: &identity::Keypair) -> Self {
+        let ed25519_keypair = match keypair {
```
mgrand, line 106: A better way of structuring this is to have Keypair be an enum and put the logic for operations specified to the type of keypair in the impl for the keypair. If you want to see an example, look at what I do with Hash in artifact manager.

mgrand (commented):
It is difficult to give any high-level comments, because I do not know what you are trying to accomplish. It would be very helpful to have in the PR a link to a document that explains what you are trying to accomplish overall. If you have not written such a document (it doesn't need to be long), you may find the writing it will help clarify your own thoughts.

pyrsia_node/src/main.rs
prince-chrismc: There's only three lines of diff here, feel free to open a separate issue 👍

src/block_chain/block_chain.rs
```diff
@@ -13,91 +13,157 @@
...
         keypair: &identity::ed25519::Keypair,
         transactions: &[Transaction],
     ) {
-        let last_block = match self.blocks.last() {
-            Some(block) => block,
-            None => {
-                Blockchain::genesis(self, keypair);
-                return;
-            }
+        let (parent_hash, previous_number) = match self.blocks.last() {
```
prince-chrismc, line 122: You can just add the genesis block when you initialize the vector [here](https://github.com/pyrsia/pyrsia/pull/295/files#diff-ad2b62fe242505ea5bceca51cd0d423999a3aa8de6515c426aa166aff4d9a505R112-R113) and just assert the size is >= 1

src/block_chain/block_chain.rs
prince-chrismc: It's hard to image Pyrsia having more then one blockchain at this time 🤔

Cargo.lock
```diff
@@ -3642,6 +3642,7 @@ dependencies = [
 [[package]]
 name = "sha1"
 version = "0.6.1"
+
```
prince-chrismc, line 3645: Seem out of place 🤔

src/block_chain/block.rs
```diff
@@ -151,9 +151,49 @@ pub fn verify(pubkey: &identity::ed25519::PublicKey, msg: &[u8], sig: &[u8]) ->
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rand::Rng;
+
+    #[test]
+    fn test_build_block() -> Result<(), String> {
+        let keypair = identity::ed25519::Keypair::generate();
```
prince-chrismc, line 173: You expose a helper [here](https://github.com/pyrsia/pyrsia/pull/295/files#diff-b34f72ca0859bced09bf9b55c159ec83f584736180ab108626113dbc22072af5R154) but don't use it... was that intentional?

prince-chrismc (dismissed):
I like the idea of the config being saved in the genesis block.

Bunch of editorial comments from me but this is enough of an improvement we can save refactoring for the next PR.

Lots of good comments and questions from the team so try answering a few 🙏

pyrsia_node/src/main.rs
fishseabowl: Sure. I would refactor communication methods in the next PR.

src/block_chain/block.rs
fishseabowl: Thanks for your comments. This is the temporary solution. I would bind these methods with related struct.

src/block_chain/block.rs
fishseabowl: Sorry. What are your concerns? Display trait or unit test?

src/block_chain/block_chain.rs
fishseabowl: Test networks, different consensus algorithms, or different contents in the ledger.

src/block_chain/block_chain.rs
fishseabowl: It is currently used for easy serialization and deserialization. I would continue to refactor these codes.

src/block_chain/block_chain.rs
fishseabowl: In my definition, a genesis block is not a block. There are different data structs.

Cargo.lock
fishseabowl: Let me check. I don't use sha1 algorithm.

mgrand: Sha1 is used by Synapse for Bittorrent. V1 torrents use sha1. V2 supports SHA256. Neither Synapse nor any of the Rust based bittorrent implementations that I have found support v2. That is something to be addressed in the future. We will be compensating by having the artifact manager rehash the files using a stronger hash before allowing them to be pulled.

prince-chrismc (approved):

src/block_chain/block.rs
prince-chrismc: Just some odd code duplication/unused, nothing critical

src/block_chain/block_chain.rs
prince-chrismc: Thats very interesting. 🤔

chb0github (approved):

PR #299, opened 2022-01-28
readme clarifications

jbaruch:


readme.md
```diff
@@ -32,8 +32,10 @@ For now Pyrsia only supports Docker artifacts. Follow these steps to run a Pyrsi
 2. Clone this repo `git clone https://github.com/pyrsia/pyrsia.git`
 3. `cd pyrsia/pyrsia_node`
 4. You have 2 options:
-   - regular build: `cargo run`
+   - local build: `cargo run`
    - build in docker: `docker-compose up --build`
+
+Unless you're building on Linux platforms, please use __docker build__ only.
```
mgrand, line 38: Why?  There is no apparent problem with building for native execution on Mac.  I expect we will also develop an understanding of how to build on Windows. I think that the better way to organize this is to have build directions that are specific to the target environment. 

In the case that we are building on a different environment than the target, using `docker build` should be our recommendation since it is much simpler that cross-compilation.

mgrand (approved):

readme.md
jbaruch: So, what should be the guidance which type of build should be used when?

mgrand: @jbaruch If we are doing a native build (on a Mac for a Mac, on Linux for Linux, on Windows for Windows, ...) use  the cargo build. If we are building for a different platform (on Mac for Linux, on Windows for Linux, on Linux for Mac, ...) use a docker build with a container that is running the target OS. That way, inside of the docker container, we just do a cargo build and it runs is if it were a native build. 

The alternative is to use cargo to generate an executable for for a different platform than it is running on. That is called cross-compilation. Cross compilation requires careful configuration that is easy to get wrong. Once you get that right, you need to have the right non-native linker to build the executable. It has to be supplied with the right libraries for the target platform. These are many things that have to be exactly right. For these reasons, we do not recommend the cross-compilation route.

jbaruch: Got it. Let's merge this one and I'll run another edit to explain that.

PR #301, opened 2022-01-28
Docker manifest integration

### LINKED ISSUE #213: Enable the docker integration to use ProjectVersion structs from the metadata manager to retrieve manifests from the artifact manager.

MitaliBo:
docker v2 manifest integration with artifact manager e2e
removal of /tmp file system for docker v2 manifest files
integration of node_manager put_artifact api
added and improved existing use cases

closes #213


src/docker/v2/handlers/manifests.rs
```diff
@@ -16,64 +16,95 @@
...
-        return Err(warp::reject::custom(RegistryError {
-            code: RegistryErrorCode::ManifestUnknown,
-        }));
-    }
+        Err(_error) => {
+            debug!("Getting manifest from dockerhub and storing in pyrsia storage.");
+
+            let hash = get_manifest_from_docker_hub(&name, &tag).await?;
+            manifest_content =
+                get_artifact(hex::decode(hash).unwrap().as_ref(), HashAlgorithm::SHA512).map_err(
```
mgrand, line 90: I expected that this code would be doing what the debug message says: storing the manifest in the artifact manager. Instead it looks like it is trying to get it from the artifact manager after it get it from docker hub.

src/docker/v2/handlers/manifests.rs
```diff
@@ -16,64 +16,95 @@
...
+                    |_| {
+                        warp::reject::custom(RegistryError {
+                            code: RegistryErrorCode::ManifestUnknown,
+                        })
+                    },
+                )?;
+        }
+    };
```
mgrand, line 98: This function is large enough that it should be broken into sub-functions.

src/docker/v2/handlers/manifests.rs
```diff
@@ -259,7 +290,7 @@ fn store_manifest_in_filesystem(
     }
 
     Ok(hash)
-}
+} */
```
mgrand, line 293: Instead of commenting out large blocks of code, please delete them. If we want to see what was there, git will show it to us.

mgrand (commented):
Please see my comments

src/docker/v2/handlers/manifests.rs
```diff
@@ -16,64 +16,95 @@
...
+    let mut manifest_content = Vec::new();
+    let package_version: PackageVersion;
+    debug!("Fetching manifest for {} with tag: {}", name, tag);
+
+    //get package_version from metadata
+    match METADATA_MGR.get_package_version(DOCKER_NAMESPACE_ID, &name, &tag) {
+        Ok(pv) => {
+            if pv.is_some() {
+                package_version = pv.unwrap();
+                if !package_version.artifacts()[0].hash().is_empty() {
```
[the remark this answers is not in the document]
MitaliBo: since you mentioned in our last discussion that we can get at first index so i follow the same pattern but if that is not the case i can make this change

src/docker/v2/handlers/manifests.rs
```diff
@@ -16,64 +16,95 @@
 
 extern crate easy_hasher;
 
-use crate::artifact_manager;
-use crate::artifact_manager::HashAlgorithm;
+use super::HashAlgorithm;
+use super::*;
```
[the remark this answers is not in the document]
MitaliBo: handled

src/docker/v2/handlers.rs
```diff
@@ -16,5 +16,6 @@
 
 pub mod blobs;
 pub mod manifests;
-pub use crate::artifact_manager::HashAlgorithm;
+pub use crate::artifact_manager::*;
```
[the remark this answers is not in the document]
MitaliBo: yeah i was using it before couple of things but later i optimized the code but forgot to revert the import

src/docker/v2/handlers/blobs.rs
```diff
@@ -269,6 +269,7 @@ fn store_blob_in_filesystem(
     let push_result = put_artifact(
         hex::decode(&digest.get(7..).unwrap()).unwrap().as_ref(),
         Box::new(reader),
+        HashAlgorithm::SHA256,
```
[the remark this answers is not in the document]
MitaliBo: we do have unit test for pu_artifact in node_manager module but we do not have any test cases for blobs module, for that i have created #307 as this is different effort

src/docker/v2/handlers/manifests.rs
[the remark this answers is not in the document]
mgrand: It is true now. When you write code that relies on this, it creates a hidden dependency. This makes the code more brittle and easier to break. 

In the future, if someone changes the code in a way causes the manifest not to be the first artifact in the array, they will have no idea that they are breaking anything. 

It is clear from reading the code that every artifact is assigned a mime type. It is reasonable for a future maintainer to assume that since we have taken the trouble to assign the mime types, we will rely on them.

src/docker/v2/handlers/manifests.rs
```diff
@@ -16,64 +16,93 @@
...
+    };
 
-    let content = blob_content.unwrap();
     Ok(warp::http::response::Builder::new()
         .header(
             "Content-Type",
             "application/vnd.docker.distribution.manifest.v2+json",
         )
-        .header("Content-Length", content.len())
+        .header("Content-Length", manifest_content.len())
```
chb0github, line 103: You should never have to set this

mgrand: The framework will set it more reliably than if we do it ourselves

src/docker/v2/handlers/manifests.rs
MitaliBo: yes it is....in "get_manifest_from_docker_hub" its getting from dockerhub and storing in pyrsia and its only getting artifact after that to return back to user

src/docker/v2/handlers/manifests.rs
MitaliBo: i have not changed anything there on my side , it was set before not sure what was the intent but i will remove it

src/docker/v2/handlers/blobs.rs
```diff
@@ -269,6 +269,7 @@ fn store_blob_in_filesystem(
     let push_result = put_artifact(
         hex::decode(&digest.get(7..).unwrap()).unwrap().as_ref(),
```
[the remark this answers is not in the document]
MitaliBo: since this is about manifest i do not want to make any changes for blobs, i am adding this to be handled in #307

MitaliBo:
@mgrand i have taken care of most of the comments, please re-review

mgrand (approved):

betarelease:
As we talked about this in Slack, before merge can you squash this PR into meaningful commits so that the commit messages are useful and readable.

src/node_manager/handlers.rs
```diff
@@ -159,8 +160,12 @@ mod tests {
     fn put_and_get_artifact_test() -> Result<(), anyhow::Error> {
         debug!("put_and_get_artifact_test started !!");
         //put the artifact
-        put_artifact(&GOOD_ART_HASH, Box::new(get_file_reader()?))
-            .context("Error from put_artifact")?;
+        put_artifact(
+            &GOOD_ART_HASH,
```
betarelease, line 164: please use complete words instead of shortening - also try to describe what GOOD means here
For example a good alternative could be
ARTIFACT_WITH_VALID_HASH

src/node_manager/handlers.rs
```diff
@@ -91,8 +91,9 @@ pub fn get_artifact(
 pub fn put_artifact(
     artifact_hash: &[u8],
     art_reader: Box<dyn Read>,
+    art_algorithm: HashAlgorithm,
 ) -> Result<bool, anyhow::Error> {
-    let hash = Hash::new(HashAlgorithm::SHA256, artifact_hash)?;
+    let hash = Hash::new(art_algorithm, artifact_hash)?;
```
betarelease, line 96: can we rename this to just algorithm instead of art_algorithm (also as I read this it is not an artifact algorithm so that is wrong name anyway)?

We are specifying an algorithm for hashing so the variable name should reflect that fact instead of what it manipulates.

src/docker/v2/handlers/manifests.rs
```diff
@@ -783,8 +840,9 @@ mod tests {
 
     fn check_artifact_manager_side_effects() -> Result<(), Box<dyn StdError>> {
```
betarelease, line 841: I have not read this code before but this is not a good pattern. We are in a functional programming language so having side_effects signifies bad design. Can we fix this in this PR if not let's create an issue for removing side effects?

src/docker/v2/handlers/manifests.rs
```diff
@@ -760,20 +771,66 @@ mod tests {
...
 
     fn check_package_version_metadata() -> anyhow::Result<()> {
         let some_package_version =
             METADATA_MGR.get_package_version(DOCKER_NAMESPACE_ID, "hello-world", "v3.1")?;
         assert!(some_package_version.is_some());
+        assert_eq!("v3.1", some_package_version.unwrap().version());
         Ok(())
     }
 
+    fn check_get_manifest_result(result: Result<impl Reply, Rejection>) {
```
betarelease, line 807: can we rename this test as follows:
verify_get_manifest_result_has_expected_content_length so that it clearly communicates what is being tested?

mgrand:
@betarelease 
> As we talked about this in Slack, before merge can you squash this PR into meaningful commits so that the commit messages are useful and readable.
I usually make frequent commits, sometime just because time has passed and I want to be able to look at my work. There is not always a meaningful comment to be made. I might just be committing because I am breaking for lunch.

Do we need to stop using git in this way?  Should we only make commits when we have reached at state that we can articulate in a way that will be clear to others? This will make using git less effective.

src/docker/v2/handlers/manifests.rs
```diff
@@ -760,20 +771,66 @@ mod tests {
         check_package_version_metadata()?;
         Ok(())
     }
+    #[test]
+    fn test_handle_get_manifest() -> Result<(), Box<dyn StdError>> {
```
betarelease, line 775: can we rename this test to describe the expected result:
test_handle_get_manifest_returns_(expected manifest with properties )

src/docker/v2/handlers/manifests.rs
```diff
@@ -297,17 +257,67 @@ async fn get_manifest_from_docker_hub_with_token(
...
+        }
+        Err(error) => warn!("Error storing manifest in artifact_manager {}", error),
+    };
+    Ok(hash)
+}
+
+fn get_artifact_manifest(artifacts: &[Artifact]) -> Option<&Artifact> {
+    for artifact in artifacts {
+        if let Some(mime_type) = artifact.mime_type() {
+            debug!("mime type is: {}", mime_type);
```
betarelease, line 304: Since the code now works, (since there is a PR) can we remove the debug lines?

src/docker/v2/handlers/manifests.rs
```diff
@@ -297,17 +257,67 @@ async fn get_manifest_from_docker_hub_with_token(
     let bytes = response.bytes().await.map_err(RegistryError::from)?;
 
     debug!("Storing manifest pulled from dockerHub in artifact manager");
```
betarelease, line 259: This debug is providing the same information as the info on line 264. Can we remove this?

src/docker/v2/handlers/manifests.rs
```diff
@@ -16,64 +16,99 @@
...
 use reqwest::{header, Client};
 use serde_json::{json, Map, Value};
 use std::fmt::Display;
-use std::fs;
 use uuid::Uuid;
 use warp::http::StatusCode;
 use warp::{Rejection, Reply};
 
 // Handles GET endpoint documented at https://docs.docker.com/registry/spec/api/#manifest
 pub async fn handle_get_manifests(name: String, tag: String) -> Result<impl Reply, Rejection> {
```
betarelease, line 42: Since this is being changed and that is an opportunity to improve code readability:

This function name has repeated words - 
this is in handlers/manifests.rs
this is a handler for manifests so we can skip handle and manifests from the name and just stick to get. If get is reserved or does not describe intent we should look at similar words to explain what it is actually doing.

src/docker/v2/handlers/manifests.rs
MitaliBo: question for @mgrand as this is not my code

MitaliBo:
@betarelease i see all your comments are related to renaming and some of them in the code which has not been changed in this PR, so i think its better i address them in different PR , the one which is following this one #312 and this one is blocking my next one too.

src/docker/v2/handlers/manifests.rs
mgrand: @betarelease The side effect that we are checking is that the manifest is stored in the artifact manager. It is the job of the artifact manager to be stateful in this way and store things. Those are external side effects that are independent of the programming language.  There are many other things we do that have side effects such as logging. If you feel that external side-effects are an issue, please schedule a meeting to discuss, since there is much of that.

Regarding rust being a functional language, it does not support a purely functional style. The simple act of making an assignment can change the lifetime of a value. In some cases it is necessary to make an assignment, otherwise the compiler will complain that a value does not live long enough. That is the reason that none of us are programming in a purely functional style.

betarelease:
Yes for the comments that do pertain to old code I do not expect it to be fixed since this PR is pretty involved already.(Although I would like to see that change in the future where we refactor and clean up stuff around the code we are working on).

For all the comments I made that were lines changed in this code, I think those should be addressed as part of this PR.

If this does indeed block another PR, we need to consider that as required waiting time, since reviewers will need time(days) to review and that is just part of the process.

betarelease:
(I did not find a way to respond to this inline so I am copying the conversation here to preserve context)

@mgrand 
Do we need to stop using git in this way? Should we only make commits when we have reached at state that we can articulate in a way that will be clear to others? This will make using git less effective.

Using git to make frequent commits helps with preserving work and moving forward. Having said that, before we create a PR I expect the committer to break that PR into 3-4 meaningful commits(squash them locally, cleanup commit messages so that they read well to the users, add more context etc.) and then submit a PR.

prince-chrismc:
> As we talked about this in Slack, before merge can you squash this PR into meaningful commits so that the commit messages are useful and readable.

@betarelease This is the configuration I have forced on everyone since the beginning.

There is no need to take any manual action.

prince-chrismc (approved):

src/docker/v2/handlers/manifests.rs
MitaliBo: verify_get_manifest is better as we are checking overall result, whether its status if 200 and comparing content length

PR #303, opened 2022-01-31
Torrent on dht

mgrand:
This pull request adds logic to add torrents created by the artifact manager to the DHT (distributed hash table). The new logic uses the artifact hash as a multihash as the key to find the torrent in the DHT. 

This PR does not include a test to verify that the torrent has been added. This requires an enhancement to the `network::behavior` module which will be in the next pull request. Then enhancement will be to allow an test to be aware of the results of a search in the DHT for the torrent. The results of the search are delivered asynchronously to the behaviour module, so I will need a way for asynchronous code to get the results of Kademlia operations.

Also included is the way that the peer_id and Kademlia are initialized. They need to be initialized before the statically initialized artifact manager. 

Another change is that Kademlia is access though a thread-safe proxy that uses a mutex to ensure only one thread at a time is calling it. The Kademlia struct has multiple methods that modify the struct. For this reason, it cannot be shared between threads without the mutex.

There are also some additional unit tests that I have added to the artifact manager to increase the code coverage.


src/artifact_manager.rs
```diff
@@ -714,11 +828,22 @@ mod tests {
...
+        Ok(())
+    }
+
     #[test]
     pub fn push_artifact_then_pull_it() -> Result<(), anyhow::Error> {
         let mut string_reader = StringReader::new(TEST_ARTIFACT_DATA);
         let hash = Hash::new(HashAlgorithm::SHA256, &TEST_ARTIFACT_HASH)?;
-        let dir_name = create_tmp_dir("tmp")?;
+        let dir_name = create_tmp_dir("tmpP")?;
         let am: ArtifactManager =
```
MitaliBo, line 847: we had decided to use "Tempfile" to create these temp directories or files which automatically takes care of removing them, can you please update the test cases accordingly

mgrand: I tried changing the code to use temp_dir.  The way that it works on a Mac breaks the building of torrents with this message from the library I am using:
```
    Error creating torrent for /private/var/folders/ht/56b7cbd16t9b5ymqdxxv1rm40000gn/T/.tmpoBEImI/SHA256/6b29f2f1e5024c419506e9503e024b3d8a5a08b6f6d55b6888667952d1041554.file: failed to build torrent: Root path contains hidden components.
```
I think that it is a reasonable requirement from the library that torrents should not be in a hidden directory. I could change the tests that temp_dir does not break, but this is likely to cause future problems for the same reason if we integrate with other things.  I am going to leave it as is.

I will add comments explaining this.

efrisch (approved):
👍

MitaliBo (approved):

PR #312, opened 2022-02-02
Test cases for fetch manifest 

### LINKED ISSUE #311: Test coverage for manifest module
Some test cases have been added to manifest module but those are not enough, this is to add more coverage

MitaliBo:
closes #311 

1) Added test cases for fetch manifest if it is not available in pyrsia storage.
2) Added tokio test lib to test async functions
3) Also handled some of the comments mentioned in #301


src/docker/v2/handlers/manifests.rs
```diff
@@ -804,7 +820,26 @@ mod tests {
...
 
-    fn check_get_manifest_result(result: Result<impl Reply, Rejection>) {
+    fn verify_fetch_manifest_result_if_not_in_pyrsia(result: Result<impl Reply, Rejection>) {
+        match result {
+            Ok(reply) => {
+                let response = reply.into_response();
+                assert_eq!(response.status(), 200);
+
+                let mut headers = HeaderMap::new();
+                headers.insert("content-length", "528".parse().unwrap());
```
fishseabowl, line 830: Is "content-length"  a fixed value of 528? And it won't change in the future?
Is headers.insert("content-length", "528".parse::<u32>().unwrap()); better?

MitaliBo: 1. yes for this test case we are using a particular digest for which it will be same
2. since here we are not annotating with any datatype so both means same

fishseabowl (dismissed):
Thanks for the quick reply.

src/docker/v2/handlers/manifests.rs
```diff
@@ -756,6 +755,12 @@ mod tests {
...
 }"##;
 
+    macro_rules! test_async {
+        ($e:expr) => {
+            tokio_test::block_on($e)
+        };
+    }
+
     fn tear_down() {
         if Path::new(&env::var("PYRSIA_ARTIFACT_PATH").unwrap()).exists() {
```
efrisch, line 765: Not a priority, but I think we should declare constants `PYRSIA_ARTIFACT_PATH` and `DEV_MODE` somewhere instead of repeating the keys `"PYRSIA_ARTIFACT_PATH"` and `"DEV_MODE"` everywhere.

src/docker/v2/handlers/manifests.rs
```diff
@@ -800,27 +805,45 @@ mod tests {
...
+    fn test_fetch_manifest() {
         let name = "httpbin";
         let reference = "v2.4";
 
         let future = async {
-            handle_put_manifest(
+            put_manifest(
                 name.to_string(),
                 reference.to_string(),
                 Bytes::from(MANIFEST_V1_JSON.as_bytes()),
```
efrisch, line 816: 😮

efrisch (approved):
:+1 lgtm

erwin1 (approved):

Issue #322, opened 2022-02-03
Merge AlephBFT into Blockchain sample node


prince-chrismc:
For our hackathon we started: https://github.com/prince-chrismc/pyrsia-sandbox/tree/aleph-substrate-node

This needs to be rebased and moved to the example + library so we can get starting with making improvements

- [x] Seperate network code #332 
- [x] Align signature struct #333 
- [x] Hasher Trait #421 
- [x] Authority Pen #432
- [x] Authority Verifier and Key Verifier #439
- [x] KeyBox implementation #464 
- [x] Add codec support #535
- [ ] Network and Chain Session #600 

***

KeyBox requires index < count so the index can not be the peer id. each node will have it's own mapping 😱

prince-chrismc:
## Criteria

- have all the code compile in this repository
- example should do consensus over AlephBFT

prince-chrismc:
This is blocked by https://github.com/pyrsia/pyrsia/issues/320

PR #323, opened 2022-02-04
Move Blockchain Definition and related methods to new crate

fishseabowl:
1. Add header, block, and blockchain definition and related methods
2. Add read, add block to file functions


src/blockchain/Cargo.toml
```diff
@@ -4,7 +4,19 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ed25519-dalek = "1.0.1"
+multihash = {version = "0.15.0", features = ["serde-codec"]}
+primitive-types = "0.10.1"
+bincode = "1.3.3"
+serde = { version = "1.0", features = ["derive"] }
+#libp2p = {git = "https://github.com/libp2p/rust-libp2p.git", rev = "e15dad5f45a9ae804b7a27bd6f885a1b00e0525a", features=["tcp-tokio"]}
```
prince-chrismc, line 11:
```suggestion
```

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+
+    let local_id = pyrsia_blockchain_network::header::hash(
+        &pyrsia_blockchain_network::block::get_publickey_from_keypair(&ed25519_keypair).encode(),
+    );
+    // Kick it off
+    loop {
+        tokio::select! {
+            line = stdin.next_line() => {
+                let line = line?.expect("stdin closed");
+                let transaction = pyrsia_blockchain_network::block::Transaction::new(
```
chb0github, line 179: Can we import this and reduce the name spacing

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+                    ),
+                    &ed25519_keypair,
+                );
+                transactions.push(transaction);
+                let (parent_hash, previous_number, previous_commiter)=read_last_block(filepath.clone());
+                //let parent_hash=header::hash(b"");
+                //let previous_number=0;
+
+                if check_number==APART_ONE_COMMIT
+                && previous_commiter == local_id{
```
chb0github, line 194: Weird newline

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+        }
+    }
+}
+
+pub fn append_genesis_block(path: String, key: &identity::ed25519::Keypair) {
+    use pyrsia_blockchain_network::blockchain::GenesisBlock;
+    use std::fs::OpenOptions;
+    use std::io::Write;
+
+    let g_block = GenesisBlock::new(key);
```
chb0github, line 222: I don't think you should be treating the genesis block as exceptional. It's a normal block with hardcoded values

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+    u128,
+    pyrsia_blockchain_network::header::Address,
+) {
+    use std::io::{BufRead, BufReader};
+    let file = std::fs::File::open(path).unwrap();
+
+    let buffered = BufReader::new(file);
+
+    let line = match buffered.lines().last() {
+        Some(v) => v,
```
chb0github, line 245: You could just use `expect` here and shortcut the match

src/blockchain/src/block.rs
```diff
@@ -0,0 +1,160 @@
...
+pub fn get_publickey_from_keypair(
+    keypair: &identity::ed25519::Keypair,
+) -> identity::ed25519::PublicKey {
+    keypair.public()
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Block {
+    pub header: Header,
+    pub transactions: Vec<Transaction>,
```
chb0github, line 56: Is there something like a `SortedSet` in rust? If we use a sorted set then these transactions will naturally stay sorted.

src/blockchain/src/block.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn verify(&self) -> bool {
+        let pubkey = match libp2p::identity::ed25519::PublicKey::decode(&self.signature.pubkey) {
+            Ok(v) => v,
+            Err(e) => {
+                println!("{}", e);
+                return false;
+            }
+        };
+
+        verify(
```
chb0github, line 85: You have 1 verify function call another? I don't get it

src/blockchain/src/blockchain.rs
```diff
@@ -0,0 +1,192 @@
...
+            blockchain_id: BlockchainId::Pyrsia,
+            hash_algorithm: HashAlgorithm::Keccak,
+            hash_size: 32, //256bits
+            signature_algorithm: SignatureAlgorithm::Ed25519,
+            key_size: 32, //256bits
+        }
+    }
+}
+
+impl Default for Config {
```
chb0github, line 63: This is interesting - what does this do?

src/blockchain/src/blockchain.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Define Genesis Block
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct GenesisBlock {
```
chb0github, line 71: We don't need a callout block for genesis

src/blockchain/src/blockchain.rs
```diff
@@ -0,0 +1,192 @@
...
+            header,
+            config,
+            signature: Signature::new(&bincode::serialize(&header.current_hash).unwrap(), keypair),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Blockchain {
+    pub genesis_block: GenesisBlock,
```
chb0github, line 99: There are just blocks

src/blockchain/src/header.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq)]
+pub struct HashDigest {
+    multihash: Multihash,
+}
+
+pub type Address = HashDigest;
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy)]
+pub struct Header {
```
prince-chrismc, line 29: You forgot to remove the old code 

https://github.com/pyrsia/pyrsia/blob/385d406acf217d27ea9939a98a6a670084cadecc/src/block_chain/header.rs

chb0github (changes requested):
I have a few comments. We want this to be our "clean" impl. People will look

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+    // `TokioTcpConfig` is available through the `tcp-tokio` feature.
+    tcp::TokioTcpConfig,
+    Multiaddr,
+    NetworkBehaviour,
+    PeerId,
+    Transport,
+};
+use rand::Rng;
+use std::error::Error;
+use tokio::io::{self, AsyncBufReadExt};
```
prince-chrismc, line 39: Following christians idea we can skip this name

```suggestion
use tokio::io::{self, AsyncBufReadExt};
use pyrsia_blockchain_network::*;
```

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+pub const APART_ONE_COMMIT: &str = "2";
+
+/// The `tokio::main` attribute sets up a tokio runtime.
 #[tokio::main]
-async fn main() {
-    pretty_env_logger::init();
-    let _e: Example = 0;
+async fn main() -> Result<(), Box<dyn Error>> {
+    // Create a random PeerId
+    let id_keys = pyrsia_blockchain_network::blockchain::generate_ed25519();
```
prince-chrismc, line 49:
```suggestion
    let id_keys = blockchain::generate_ed25519();
```

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+    let ed25519_keypair = match id_keys {
+        identity::Keypair::Ed25519(v) => v,
+        identity::Keypair::Rsa(_) => todo!(),
+        identity::Keypair::Secp256k1(_) => todo!(),
+    };
+
+    let mut transactions = vec![];
+
+    let local_id = pyrsia_blockchain_network::header::hash(
+        &pyrsia_blockchain_network::block::get_publickey_from_keypair(&ed25519_keypair).encode(),
```
prince-chrismc, line 172:
```suggestion
    let local_id = header::hash(
        &block::get_publickey_from_keypair(&ed25519_keypair).encode(),
```

prince-chrismc (changes requested):
Blocker: We need to remove the old code (since most of these files are copies)

Please try to remove the `pyrsia_blockchain_network` from the name to keep it clean

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+    // Create a Floodsub topic
+    let floodsub_topic = floodsub::Topic::new("block");
+
+    // We create a custom network behaviour that combines floodsub and mDNS.
+    // The derive generates a delegating `NetworkBehaviour` impl which in turn
+    // requires the implementations of `NetworkBehaviourEventProcess` for
+    // the events of each behaviour.
+    #[derive(NetworkBehaviour)]
+    #[behaviour(event_process = true)]
+    struct MyBehaviour {
```
prince-chrismc, line 80: I'd like to give this a better name, maybe

```suggestion
    struct BlockchainNetworkBehaviour {
```

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+        floodsub: Floodsub,
+        mdns: Mdns,
+    }
+
+    impl NetworkBehaviourEventProcess<FloodsubEvent> for MyBehaviour {
```
prince-chrismc, line 85: It would be ideal to move this to a seperate file so we have less conflicts when we do #322 

I have a lot grouped into https://github.com/prince-chrismc/pyrsia-sandbox/blob/aleph-substrate-node/src/block_chain/network.rs which might be something we can do before

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+        // Called when `floodsub` produces an event.
+        fn inject_event(&mut self, message: FloodsubEvent) {
+            if let FloodsubEvent::Message(message) = message {
+                //println!("Received: '{:?}' from {:?}", &message.data, message.source);
+                let block: pyrsia_blockchain_network::block::Block =
+                    bincode::deserialize::<pyrsia_blockchain_network::block::Block>(&message.data)
+                        .unwrap();
+                println!("++++++++++");
```
prince-chrismc, line 93: We should replace these with the logging library we have used in the rest of the project!

src/blockchain/examples/simple_node.rs
```diff
@@ -18,10 +18,255 @@ extern crate pretty_env_logger;
...
+        println!("Dialed {:?}", to_dial)
+    }
+
+    // Read full lines from stdin
+    let mut stdin = io::BufReader::new(io::stdin()).lines();
+
+    // Listen on all interfaces and whatever port the OS assigns
+    swarm.listen_on("/ip4/0.0.0.0/tcp/0".parse()?)?;
+
+    let ed25519_keypair = match id_keys {
```
prince-chrismc, line 163: We should only focus on one key. Directly using `identity::Keypair::Ed25519` we can refactor later to use the generic enum version

Thoughts?

src/blockchain/examples/simple_node.rs
prince-chrismc: > I don't think you should be treating the genesis block as exceptional. It's a normal block with hardcoded values

This is something we can explore when we do #322 👍

src/blockchain/src/block.rs
```diff
@@ -0,0 +1,160 @@
...
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+use std::time::{SystemTime, UNIX_EPOCH};
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum TransactionType {
+    Create,
+}
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Signature {
+    signature: Vec<u8>,
```
prince-chrismc, line 27: It would be nice if we can use the edwards key library like I had done https://github.com/prince-chrismc/pyrsia-sandbox/blob/a9333aced792e1017db1f424eb1b4b77c7803aa4/src/block_chain/crypto.rs#L28

This way there will be less conflicts for #322

prince-chrismc (commented):
# Ideas for future work

I'll open these as issues when this is merged!

src/blockchain/examples/simple_node.rs
chb0github: Go for it

src/blockchain/examples/simple_node.rs
fishseabowl: Good comments. I would update it.

prince-chrismc (approved):
My blockers have been addressed 🚀

src/blockchain/src/blockchain.rs
prince-chrismc: This is one of the trendy rust patterns that exists... instead of having a ctor

https://rust-unofficial.github.io/patterns/idioms/default.html

src/blockchain/examples/simple_node.rs
fishseabowl: Absolutely. I would add log and debug information.

src/blockchain/examples/simple_node.rs
fishseabowl: Yes. It is just for backward compatible with Pyrsia

src/blockchain/examples/simple_node.rs
fishseabowl: I store the configuration information in the genesis block. I would refactor these codes to make them clearer.

prince-chrismc (dismissed):

src/blockchain/src/block.rs
prince-chrismc: Maybe like https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html

This is an excellent point to discuss!

Thinking a little bit, I dont think the order of transactions matters 🤔 we are looking at starting with an AURA so only one node will commit and to my knowledge other nodes wont challenge the block based on the transactions inside.

Also with Aleph there will only be a linear graph so it should be less of a problem at the block level.

Great Explore/Spike issue 👍

prince-chrismc:
There's two small comments that should be addressed in this PR

https://github.com/pyrsia/pyrsia/pull/323#discussion_r799676881
https://github.com/pyrsia/pyrsia/pull/323#discussion_r799678207

***

@chb0github, can we save https://github.com/pyrsia/pyrsia/pull/323#discussion_r799679109 for a follow up PR?
My code has a complete different the notion of the genesis block and I think it might be worth making changes with #322 since we'll need to discuss and decide what we want.

Thoughts?

src/blockchain/examples/simple_node.rs
prince-chrismc: Christian, good for following this up in an issue

He'd like to have the genesis block ad index 0 in the blockchain, maybe a special transaction?

***

Xuefei, when we have a "sync initialize" when nodes join this can go away (since only one node will make the configuration)

prince-chrismc: moved to #329

src/blockchain/src/block.rs
prince-chrismc: Error prone, can solve the duplicated transaction being recorded.

e.g the same message being routed to us twice from network

prince-chrismc: moved to #331

src/blockchain/src/block.rs
prince-chrismc: I agree this is a little awkward. but we need to change this for #322 so let's leave it alone for now 🙈

src/blockchain/examples/simple_node.rs
prince-chrismc:
```suggestion
    # TODO(prince-chrismc): Give this a much better name!
    struct MyBehaviour {
```

src/blockchain/src/block.rs
prince-chrismc: moved to #333

prince-chrismc (dismissed):

prince-chrismc:
We all approved!

image

betarelease:
This particular PR also has a lot of commits that are point in time but need to be squashed to 4-5 meaningful commits. I am not able to find a conversation that created a another ticket - since this PR is too long.

betarelease:
Also if the three of you are working together, shouldn't we wait for someone outside the three of you to approve as well?

betarelease:
Github tweeted this trick to reduce noise from merge commits

https://twitter.com/github/status/1489672242074206209?s=20&t=bBtLBVUZabSAlSrMCKWf3w

prince-chrismc:
> I am not able to find a conversation that created a another ticket - since this PR is too long.

Please follow slack, I already enabled this on the repo

https://openssf.slack.com/archives/C0302CVKPA5/p1644000250494039

src/blockchain/examples/simple_node.rs
prince-chrismc: Moved to #332

src/blockchain/examples/simple_node.rs
prince-chrismc:
```suggestion
                println!("++++++++++"); // TODO(fishseasbowl): Replace with logging methods
```

src/blockchain/examples/simple_node.rs
prince-chrismc: Moved to #333

prince-chrismc:
After a quick zoom call,

> reduce noise from merge commits

We'll make sure to not open PRs with messy, unhelpful commits so are to not pollute the reader. Team effort to help with this.

I will manually be merging this... at at that time I can fixup the message that will be made so regardless of what's in the PR it will look pretty. 👍

Issue #302, opened 2022-01-31
Configure pyrsia artifacts storage directory with correct permissions


MitaliBo:
During installation we want to have a directory named "pyrsia" setup with all the permissions which can be used for Artifacts storage and retrieval. As of current code we create "pyrsia" in current directory. This directory path should be set as env variable which can be used in code

sbtaylor15:
@MitaliBo I am assuming 644 for the permission will work.  Correct?

MitaliBo:
yes i think that would do

MitaliBo:
We have "PYRSIA_ARTIFACT_PATH=/usr/local/var/pyrsia" set and env var is exposed which now can be consumed in code.

Issue #300, opened 2022-01-28
The Directory Used by Artifact Manager Should be Configurable


mgrand:
The artifact manager currently uses `./pyrsia` as the directory for it repository. This should be configurable by an environment variable named `PYRSIA_ARTIFACT_PATH`. The current value should be the default of the environment variable is not specified.

Issue #251, opened 2022-01-21
Configure logging in the Pyrsia linux installer


erwin1:
Discovered in #244 

We should configure logging in the apt-get installed version of Pyrsia, so that a user can somehow inspect the logs if necessary.

The release binary contains log (INFO and higher) but they are only printed if the `RUST_LOG` env variable is set to `info`

- the script starting the Pyrsia node should set `RUST_LOG=info` 
- it could then redirect that output to a file?

sbtaylor15:
Using stdout and stderr will give us the most flexibility.  When running in docker stdout and stderr can be view using the `docker logs`.  Same when the container is running in Kubernetes, `kubectl logs`.  Running natively in Ubuntu under systemctl the output will be sent to syslog and viewed with `systemctl status` or `journalctl -u pyrsia.service`

erwin1:
Great, then I think simply setting `RUST_LOG=info` is all we need to do. It will then print info logs and higher to stderr.

sbtaylor15:
Ok, I will make sure that is set for the service

sbtaylor15:
Added RUST_LOG=info to pyrsia-node docker image in PR #260

Issue #129, opened 2021-12-17
Use cargo-deb to create .deb files


sbtaylor15:
Need to investigate https://crates.io/crates/cargo-deb and implement if appropriate.  At first glance it looks to have all the functionality we need.

PR #328, opened 2022-02-07
Use cargo-deb to install synapse and pyrsia under systemd

### LINKED ISSUE #314: Research what `rust-strip` does; and whether we should include it in our release build process.
Part of `cargo-binutils`, performing a release build produces a 14 MB executable. Running `rust-strip` on that reduces the size to 10 MB. How?

sbtaylor15:
- Replaced dpkg-deb with the cargo-deb action.

  cargo-deb uses key/values from the Cargo.toml as input to create the .deb file.  One of the required fields is the `version` number.  This was changed from 0.1.0 to 0.1.0+&lt;build number&gt; per the [semver.org](https://semver.org/).  The build number was added to make the new .deb file unique.  Debian repositories are set up so you cannot replace existing versions.  Instead, you must iterate the version number to add updates.   Adding the build number allows us to iterate the version number but keeps the main version pinned to 0.1.0.   

  Since the version number is in the Cargo.toml and it is being updated, --locked had to be removed as part of the build.  --offline was tested but the pyrsia/synapse is a dependency that is not cached locally but is being built as well. 

- Use cargo-strip to reduce the size of the binaries.  

  cargo-strip runs the OS level `strip` program to remove debug and unused symbols.  This is equivalent to running the linker with the `-s` parameter.

- Added pyrsia/synapse build to pyrsia/pyrsia Rust Workflow.  This was done to minimize token setup and synchronization issues across repositories.  Synapse is being built in its own Job in order to prevent errors when cloning another repository into a  workspace of pyrsia/pyrsia.

- Added service install and service configuration for Pyrsia and Synapse.

  cargo-deb will create and install the services under `systemd`.   `RUST_LOG=info` and `PYRSIA_ARTIFACT_PATH=/usr/local/var/pyrsia` environment variables are set up as well.  Synapse will startup under its default configuration.  Pyrsia install will reconfigure Synapse service to use its configuration file.  

  Directory and File Layout:
  - /usr/bin/pyrsia_node
  - /usr/bin/pyrsia
  - /usr/share/doc/pyrsia/README
  - /usr/share/pyrsia/synapse.toml
  - /usr/bin/synapse
  - /usr/share/doc/synapse/README
  - /usr/local/share/synapse/synapse.toml
  - /usr/local/share/synapse/sycli.toml

See #300
Resolves #314, #302, #251, #129

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,124 +30,120 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         brew install openssl
+
+    # Install cargo deb and strip
+    - name: Install cargo-deb cargo-strip
+      if: runner.os == 'Linux'
+      uses: actions-rs/cargo@v1.0.1
+      with: 
+        command: install
+        args: cargo-deb cargo-strip --force
+
+    # Need to add build number to version number
+    - name: Iterate Build Version Number
+      run: |
+        sed -i.bak "/^version = \"0.1.0\"/s/\"\$/+${{ github.run_number }}\"/" Cargo.toml
+
+    - name: Build Release Binaries
+      run: |
+        cargo build --workspace --verbose  --release
     
-    - run: cargo build --workspace --verbose --locked --release
-    - run: cargo test --workspace --verbose
+    - name: Execute Test Cases
+      run: |
+        cargo test --workspace --verbose
 
-    # save the pyrsia_node binary for the package creation
-    - name: Save pyrsia_node binary
-      uses: actions/upload-artifact@v2
-      if: runner.os == 'Linux'
-      with:
-        name: pyrsia_node
-        path: target/release/pyrsia_node
-        if-no-files-found: error
-
-    # save the pyrsia binary for the package creation
-    - name: Save pyrsia binary
-      uses: actions/upload-artifact@v2
-      if: runner.os == 'Linux'
-      with:
-        name: pyrsia
-        path: target/release/pyrsia
-        if-no-files-found: error
-
-    # save the debian control files to be used with the deb packager
-    - name: Save debian control file
-      uses: actions/upload-artifact@v2
-      if: runner.os == 'Linux'
-      with:
-        name: DEBIAN
-        path: installers/DEBIAN
-        if-no-files-found: error
-    
-      # build synapse-bt 1.0 from github (commit:634ee2a62 = 1.0 tag)
-    - name: Build synapse 
+    - name: Strip Binaries
       if: runner.os == 'Linux'
       run: |
-        cargo install --force --git https://github.com/Luminarys/synapse.git#634ee2a62 synapse-bt
-        cp ~/.cargo/bin/synapse target/release/synapse
-      shell: bash
+        cargo strip
 
-    # save the debian control files to be used with the deb packager
-    - name: Save synapse binary
-      uses: actions/upload-artifact@v2
+    # Create Pyrsia .deb file
+    - name: Package Pyrsia as deb file
       if: runner.os == 'Linux'
+      uses: actions-rs/cargo@v1.0.1
       with:
-        name: synapse
-        path: target/release/synapse
-        if-no-files-found: error
+        command: deb
+        args: --no-build -v
 
-  # Create the .deb file, add it to the apt repo and
+    # Add the prysia .deb file to the apt repo which is part of
+    # the github website repo
+    - name: Upload Pyrsia .deb to website repo
+      uses: sbtaylor15/apt-repo-action@v2.0.4
+      # Only when we push on the main repository should we upload the results
+      if: github.repository_owner == 'pyrsia' && github.event_name == 'push' && runner.os == 'Linux'
+      with:
+        github_token: ${{ secrets.APT }}
+        repo_supported_version: |
+          focal
+        repo_supported_arch: |
+          amd64
+        page_branch: gh-pages
+        file: /github/workspace/target/debian/pyrsia_0.1.0+${{ github.run_number }}_amd64.deb
+        file_target_version: focal
+        public_key: ${{ secrets.GPG_PUBLIC }}
+        private_key: ${{ secrets.GPG_PRIVATE }}
+        key_passphrase: ${{ secrets.GPG_SECRET }}
+        github_repository: pyrsia/pyrsia.github.io
+    
+  # Create the Synapse binaries and .deb file, add it to the apt repo and
   # push the changes to the website github repo
-  apt-package:
+  synapse-package:
       needs: [build]
       runs-on: ubuntu-latest
       steps:
+      # Install cargo deb and strip
+      - name: Install cargo-deb
+        uses: actions-rs/cargo@v1.0.1
+        with: 
+          command: install
+          args: cargo-deb cargo-strip --force
 
-      # grab the pyrsia_node from the build step
-      - name: Fetch pyrsia_node binary
-        uses: actions/download-artifact@v2
+      # Checkout Synapse
+      - name: Checkout Synapse
+        uses: actions/checkout@v2
         with:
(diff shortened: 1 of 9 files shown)


installers/debian/postinst
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+/usr/bin/systemctl stop synapse-bt
+export PW=`echo ZG9udGhhY2ttZQ== | base64 -d`
+sed -i.bak "s/hackme/$PW/" /usr/share/pyrsia/synapse.toml
```
efrisch, line 5: Be explicit with sed path? Also, `/bin/sh` is often dash. I'd prefer `/usr/bin/env bash`

efrisch (approved):
Apart from my comments about the path to sed, and ambiguity about `/bin/sh` (https://wiki.ubuntu.com/DashAsBinSh) this looks good to me. 👍

installers/debian/postinst
sbtaylor15: Added /usr/bin/env bash and full path to sed

installers/debian/preinst
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
```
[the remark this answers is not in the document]
sbtaylor15: Added `#!/usr/bin/env bash` to ensure running under bash instead of dash

.github/workflows/rust.yml
```diff
@@ -30,124 +30,120 @@ jobs:
...
-      uses: actions/upload-artifact@v2
-      if: runner.os == 'Linux'
-      with:
-        name: pyrsia
-        path: target/release/pyrsia
-        if-no-files-found: error
+    # Need to add build number to version number
+    - name: Iterate Build Version Number
+      run: |
+        sed -i.bak "/^version = \"0.1.0\"/s/\"\$/+${{ github.run_number }}\"/" Cargo.toml
```
prince-chrismc, line 45: Absolutely love this 🚀 

One small grip... run number != code and this should answer "what code is being run" 

SHAs are easier to pin down and track... so I've love to tweak this

```suggestion
        sed -i.bak "/^version = \"0.1.0\"/s/\"\$/+${{ github.sha}}\"/" Cargo.toml
```

.github/workflows/rust.yml
```diff
@@ -30,124 +30,120 @@ jobs:
...
       run: |
         brew install openssl
-    
-    - run: cargo build --workspace --verbose --locked --release
-    - run: cargo test --workspace --verbose
 
-    # save the pyrsia_node binary for the package creation
-    - name: Save pyrsia_node binary
-      uses: actions/upload-artifact@v2
+    # Install cargo deb and strip
```
prince-chrismc, line 34: same as the name

```suggestion
```

.github/workflows/rust.yml
```diff
@@ -30,124 +30,120 @@ jobs:
...
-    # save the pyrsia binary for the package creation
-    - name: Save pyrsia binary
-      uses: actions/upload-artifact@v2
-      if: runner.os == 'Linux'
-      with:
-        name: pyrsia
-        path: target/release/pyrsia
-        if-no-files-found: error
+    # Need to add build number to version number
+    - name: Iterate Build Version Number
```
prince-chrismc, line 43: Sorry for being picky... this is completely optional

```suggestion
    - name: Add Build to Version Number
```

.github/workflows/rust.yml
```diff
@@ -30,124 +30,120 @@ jobs:
...
+      run: |
+        cargo test --workspace --verbose
+
+    - name: Strip Binaries
       if: runner.os == 'Linux'
       run: |
-        cargo install --force --git https://github.com/Luminarys/synapse.git#634ee2a62 synapse-bt
-        cp ~/.cargo/bin/synapse target/release/synapse
-      shell: bash
+        cargo strip
```
prince-chrismc, line 58: Why is this being done a second time https://github.com/pyrsia/pyrsia/pull/328/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR40

It's not obvious in the diff sorry

.github/workflows/rust.yml
```diff
@@ -30,124 +30,120 @@ jobs:
...
+      # Only when we push on the main repository should we upload the results
+      if: github.repository_owner == 'pyrsia' && github.event_name == 'push' && runner.os == 'Linux'
+      with:
+        github_token: ${{ secrets.APT }}
+        repo_supported_version: |
+          focal
+        repo_supported_arch: |
+          amd64
+        page_branch: gh-pages
+        file: /github/workspace/target/debian/pyrsia_0.1.0+${{ github.run_number }}_amd64.deb
```
prince-chrismc, line 81:
```suggestion
        file: /github/workspace/target/debian/pyrsia_0.1.0+${{ github.sha }}_amd64.deb
```

prince-chrismc (commented):
This PR is really really big and tackles a few issues in one shot

Are you open to the idea of breaking it down a little bit? 

I was not able to get through all of it 😢 but here are my initial comments

sbtaylor15:
> > * Since the version number is in the Cargo.toml and it is being updated, --locked had to be removed as part of the build
> 
> Looking quickly online I see https://doc.rust-lang.org/cargo/commands/cargo-update.html#update-options and comments saying you can do this yo your own crate.
> 
> This might be a worth test to see if we can keep the stability of `--locked` versions

@prince-chrismc - I did a test using --workspace and --locked to just update the Cargo.lock with the new version number.  You get an error with this combination.  Dropping the --locked and just using --workspace does seem to only update the version number in the Cargo.lock

.github/workflows/rust.yml
sbtaylor15: The build portion of the version number needs to be sequential.  That is how the apt-get installer determines if it needs to do an upgrade of a package.  The sha is not sequential.

.github/workflows/rust.yml
sbtaylor15: > 

Many changes are needed for it all to work together.  I can break it into smaller parts but the builds will probably be broken until all the changes are in place.

.github/workflows/rust.yml
sbtaylor15: We are doing two builds.  One for Pyrsia and another for Synapse.  Initially I wanted to do Synapse under its own repo but you did not want to setup the tokens and potential introduce drift between Pyrsia and Synapse.  So doing it in one place per your comments.

.github/workflows/rust.yml
sbtaylor15: Need to build number since it is sequential.  SHA will not be sequential.

.github/workflows/rust.yml
sbtaylor15: Changed to `Add Build to Version Number`

.github/workflows/rust.yml
prince-chrismc: Ahhh thanks for explaining that differently. it makes a lot more sense now 👍

prince-chrismc:
> does seem to only update the version number in the Cargo.lock

That's unfortunate but it makes sense. Seems like this is an active area of development, found this https://github.com/crate-ci/cargo-release which is looking promising 🤔 but it's not stable

installers/debian/postinst
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+/usr/bin/systemctl stop synapse-bt
+export PW=`echo ZG9udGhhY2ttZQ== | base64 -d`
```
prince-chrismc, line 4: `donthackme` lol 😆 

Love it.

prince-chrismc (approved):
This address a ton of issues, all my questions have been answer.

Only one nit outstand but that's not required!

LGTM 🚀

PR #353, opened 2022-02-11
Replaced ART_MGR_DIR to get path value from env var

### LINKED ISSUE #302: Configure pyrsia artifacts storage directory with correct permissions

MitaliBo:
This PR closes #302 

1) Replaced ART_MGR_DIR hardcoded path to get from env variable
2) Introduced DEV_MODE env var for development and testing which will create PYRSIA_ARTIFACT_PATH
3) Fixed tests based on above changes
4) Updated readme and dockerfile to include above env vars.

### DIFF
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -12,7 +12,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y wget gnupg
           wget -qO - https://pyrsia.io/public.key | sudo apt-key add -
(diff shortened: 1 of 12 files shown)


sbtaylor15 (dismissed):
Looks good.  No changes are needed.

efrisch (approved):
lgtm. 👍

sbtaylor15 (approved):
/lgtm

Issue #347, opened 2022-02-10
Add a issue-form template


AbhijithGanesh:
Similar to #345 We need issue forms to increase dev-productivity , it'd streamline the process of making issues to the repository

### Examples

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema

AbhijithGanesh:
Can this be assigned to me ? I'll make the issue-templates, I'd also like to know what all fields and forms would be appropriate to have

mgrand:
I like the idea of issue templates as a resource that people can choose to use for some types of issues that tend to follow a structure, such as bug reports or technical debt. However, it should still be possible for people to enter free-form issues for those that do not fit an existing template.

PR #354, opened 2022-02-11
Addition of issue templates for bug reports and features

AbhijithGanesh:
Addresses #347 

Signed-off-by: Abhijith Ganesh <g.abhijith14122003@gmail.com>

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,32 @@
+name: "❌ Bug report 🐛"
+description: Report errors bugs and unexpected behavior
+labels: ["bug", "triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please make sure to [search for existing issues](https://github.com/pyrsia/pyrsia/issues) before creating a new report.
+
+        If possible, upload an image of the case as well, it'd be useful for the maintainers.
+
+  - type: textarea
+    attributes:
+      label: Steps taken/Steps to reproduce
+      placeholder: Tell us the steps required to experience the bug.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Expected Behavior
+      description: If you want to include screenshots, paste them into the markdown editor below.
+      placeholder: What were you expecting?
+    validations:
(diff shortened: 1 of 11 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,32 @@
+name: "❌ Bug report 🐛"
```
prince-chrismc, line 1: It would be nice to have more information about the platform the issue happens on

If we could always have 
- OS
- rust version
- pyrsia version (tag or sha)

That would be helpful when investigating problems

AbhijithGanesh: I agree, I'll add

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,35 @@
...
+    validations:
+      required: false
+
+  - type: textarea
+    attributes:
+      label: Experienced Instance of the bug
+      placeholder: |
+        Rust Version: 
+        Operating system:
+        Pyrsia SHA/Version :
```
prince-chrismc, line 33: Just a nit

```suggestion
        Pyrsia SHA/Version:
```

prince-chrismc (approved):

sbtaylor15 (commented):
@prince-chrismc @AbhijithGanesh  Do we need to find out if the bug is happening in the CLI or on the Node side?  Maybe optional info.  The user may or may not know.

Also, the Pyrsia Version field may cover this.  Thoughts?

prince-chrismc:
Hmmm 🤔 How will a user know the difference... the main entry point in the CLI and it should only talk to a node?
Both parts are always involved?

Maybe in the future we can circle back to add that is we see node only or cli only use cases?

AbhijithGanesh:
The separation of concerns is a good topic but I don't think the user gets relevant details about it. As our project evolves, we as devs should be able to distinguish clearly.

sbtaylor15 (approved):
Looks good.  We will separate out capturing cli vs node down the road.  As stated, they may not know were the bug is coming from.

PR #358, opened 2022-02-15
Blockchain API outline

chb0github:


prince-chrismc:
The issue was solved on the side

Issue #307, opened 2022-02-01
Test cases for blobs module


MitaliBo:
Right now we do not have any test coverage for /docker/v2/blobs module, this issue is to cover that

Move "        hex::decode(&digest.get(7..).unwrap()).unwrap().as_ref(),
" in new function

betarelease:
As part of the testing, 

these hardcoded urls should become injectable

"/tmp/registry/docker/registry/v2/repositories/{}/_manifests/tags/{}/current/link",

"/tmp/registry/docker/registry/v2/blobs/sha256/{}/{}/data",

"http://localhost:7878/v2/{}/manifests/sha256:{}",

"/tmp/registry/docker/registry/v2/repositories/{}/_uploads/{}/data",

"/tmp/registry/docker/registry/v2/repositories/{}/_uploads/{}",

These urls need to be derived from a configuration setting and the processing needs to be relative to that. The test should allow me to specify a different starting directory but still the code should work.

PR #368, opened 2022-02-22
#307 adds tests to the blobs module

betarelease:
## PR Checklist

- [x ] I've built the code `cargo build`. For major changes, `cargo build --workspace --release` is recommended.

- [x] I've read ["What is a Good PR?"](good_pr.md)

## Description

Fixes pyrsia/pyrsia#307

This PR adds tests to the blobs module and refactors the concern for handling the storage of the blob into its own file

### DIFF
--- a/src/docker/v2/models/blobs.rs
+++ b/src/docker/v2/models/blobs.rs
@@ -64,7 +64,7 @@ fn create_upload_directory(name: &str, id: &str) -> std::io::Result<()> {
     ))
 }
 
-fn store_blob_in_filesystem(
+pub fn store_blob_in_filesystem(
     name: &str,
     id: &str,
     digest: &str,
--- a/src/docker/v2/models/blobs.rs
+++ b/src/docker/v2/models/blobs.rs
@@ -1,0 +1,118 @@
+/*
+   Copyright 2021 JFrog Ltd
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


betarelease:
@MitaliBo Can you take a look at the changes I made here? I want to get this merged and follow this up with a PR so that we can optimize the download and save of the docker blobs as a different issue.

src/docker/v2/models/blobs.rs
```diff
@@ -0,0 +1,118 @@
+/*
```
MitaliBo, line 1: why was blobs code moved inside models? this is not relevant module

readme.md
```diff
@@ -92,11 +92,35 @@ For now Pyrsia only supports Docker artifacts. Follow these steps to run a Pyrsi
...
+    cargo run --bin pyrsia_node -- -p 7888
+
+    # RUST_LOG=debug cargo run --bin pyrsia_node -- -p 7888 # Use this environment variable if you would like to see debug logs
+    ```
+
+    ```
+    cargo run --bin pyrsia_node -- -p 8181
+
+    # RUST_LOG=debug cargo run --bin pyrsia_node -- -p 8181 # Use this environment variable if you would like to see debug logs
+    ```
```
MitaliBo, line 111: this file does not look updated

MitaliBo:
@betarelease i think we need to do some cleanup in this PR and this not well refactored as its just duplicated the code in another module which is actually disconnected with rest of the code. It actually does not fulfill the requirement of #307 so we should remove this with that issue otherwise that issue will get closed if this gets merged.

readme.md
betarelease: I will update this. Last time I updated was > 1 week ago

src/docker/v2/models/blobs.rs
betarelease: I think I missed removing the original code from handlers. But the idea is handlers should only do the handling of the API and model goes and does the saving.(like Model and Controller - handler is the controller in this case).

Also this helps us separate the concern of communicating with dockerhub(which is what handlers does) and storing(which is what the model does)

This provides cleaner interface for testing and independent testing. We can then swap the storage model with whatever else we decide to use(say s3 for example) or a different way of storing blobs without touching handlers.

betarelease:
closing this since I opened #431

Issue #373, opened 2022-02-23
Add cargo-make to allow for running tasks in order for ci


betarelease:
cargo make allows chaining of tasks to be run as part of frequently used flows.(https://sagiegurari.github.io/cargo-make/)

Currently submitting a PR requires us to run multiple cargo commands, cargo-make allows to automate this and ensures that they will run in the same order on all developers machines and possibly on github actions as well.

Currently cargo-make has a dependency `chrono` which is failing the audit-check. Try to work on this with a newer version of `chrono` or `cargo-make` and then build a ci_flow that works consistently everywhere.

There is no version of chrono that fixes this yet. There is an open issue for this in the chrono repo: https://github.com/chronotope/chrono/issues/499

But, since we decided to use chrono for other reasons, we can continue to use that dependency with cargo-make as well.

betarelease:
Closing this issue - since the PR is closed(since cargo-make is too heavy and has dependencies that have vulnerabilities). This is being addressed by a simple https://github.com/pyrsia/pyrsia/pull/459/

PR #374, opened 2022-02-23
#373 Adds ability to automate developer tasks and ci flows

betarelease:
- adds a command chain to run the ci process locally before commit/PR

## PR Checklist

- [x ] I've built the code `cargo build`. For major changes, `cargo build --workspace --release` is recommended.
- [ x] I've run the automated unit tests `cargo test`. This executes our automated unit tests.
- [ x] I've not broken any existing tests or functionality. In addition to `cargo test`, I've run `cargo clippy`
- [ x] I've not introduced any new known security vulnerabilities. I've run `cargo audit`

- [x] I've included a brief description and a good title of the proposed changes and how to test them.
- [x] I've read the [contributing guidelines](https://github.com/pyrsia/.github/blob/main/contributing.md).
- [x] I've associated an [issue](https://github.com/pyrsia/pyrsia/issues) with this Pull Request. 
- [x] I've checked that the code is up-to-date with the `pyrsia/main` branch.
- [x] I've read ["What is a Good PR?"](good_pr.md)
- [x] I've assigned this Pull Request to "pyrsia/collaborators"

## Description

Fixes pyrsia/pyrsia#373 
Adds ability to automate developer tasks 
- adds a command chain to run the ci process locally before commit/PR

PR #393, opened 2022-02-24
#373 Adds ability to automate developer tasks

betarelease:
- adds a command chain to run the ci process locally before commit/PR

## PR Checklist

- [x] I've built the code `cargo build`. For major changes, `cargo build --workspace --release` is recommended.

## Description

Fixes pyrsia/pyrsia#373

Adds ability to automate developer tasks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,11 +65,20 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ca972c2ea5f742bfce5687b9aef75506a764f61d37f8f649047846a9686ddb66"
+dependencies = [
+ "memchr 0.1.11",
+]
+
+[[package]]
+name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
 dependencies = [
- "memchr",
+ "memchr 2.4.1",
 ]
 
 [[package]]
@@ -78,7 +87,7 @@ version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
 dependencies = [
- "winapi",
+ "winapi 0.3.9",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


betarelease:
Since @johanvos and @prince-chrismc were the original reviewers on my previous PR I have manually added you here.

prince-chrismc:
There's even more vulnerabilities now 💥

prince-chrismc:
Same two comments

- https://github.com/pyrsia/pyrsia/pull/374#issuecomment-1049066401 (PR title)
- https://github.com/pyrsia/pyrsia/pull/374#issuecomment-1049327572 (docs)

betarelease:
I am looking into the vulnerabilities:

assay -> tempdir (so maybe assay is not a good option) - but it exists in main and does not fail for tempdir vuln
For the others as we discussed, since cargo-make is a dev-dependency it is ok to fail audit-check

I found this PR merged with cargo-audit for ignoring dev-dependencies - https://github.com/rust-lang/cargo/pull/5012. How can we turn this on for our cargo-audit?

prince-chrismc:
> I found this PR merged with cargo-audit for ignoring dev-dependencies - [rust-lang/cargo#5012](https://github.com/rust-lang/cargo/pull/5012).

Just to repeat so we are on the same page... That PR is for cargo in general not the audit tool from [RustSec](https://github.com/rustsec/rustsec) that's flagging the problem. In other words, there's "experimental" support in rust but not yet in the security tooling.

> How can we turn this on for our cargo-audit?

There's no easy way to pipe this in without it introducing some inconsistency AFAIK. Open to ideas.

➡️ here's the audit tool's behavior https://github.com/rustsec/rustsec/blob/1461a59311e93a6692decd98aa325166bfb9322e/cargo-audit/src/auditor.rs#L130-L133

We _could_ delete our lockfile and regenerate a new one (with the experimental flags) but it would contain the latest version of all the packages but given how behind we are (see #382 or #386) and some dont compile... it's not a guarantee to work.
It could also pull in a new version which is compromised, a false positive, which is more confusing.

betarelease:
You are right. My google foo is failing due to `cargo audit` and `cargo-audit` and how google does not understand the difference.

I found a relevant issue that someone raised with cargo-audit on their gihub - https://github.com/rustsec/rustsec/issues/501

betarelease:
I feel like this is really important automation we should add in some way so that everyone on the team runs the same exact steps without having to remember all the -- options in the correct order. 

`cargo-make` makes it possible and easy. How else should we move forward? I would like something that is committed to the codebase instead of in the documentation on the PR template.

betarelease:
@prince-chrismc and I discussed the concerns around `cargo-make` and how heavy it is (plus adds vulnerabilities).

Although we agreed on automation required to keep the dev/PR process simple.

As a result we will not use `cargo-make` and will close this PR. 

But will work on a simpler(maybe simple bash script) to automate the steps we all need to run before submitting a PR.

PR #446, opened 2022-03-02
Update source folder readme with more terminology and sub-crate

prince-chrismc:
For better navigation to those clicking around exploring the GitHub repository

template makes no sense. just some doc 💄


chb0github (approved):

src/readme.md
```diff
@@ -1,16 +1,19 @@
 # Pyrsia Library
 
-Principale set of module making up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli)
+Principale set of module which make up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli) components.
 
 ## Terminology
```
prince-chrismc, line 5: @erwin1 @MitaliBo Do you have any more suggestions for what we could add here? Anything p2p or cli specific?

I'd like to start a list and move it soon 🤞

MitaliBo: @prince-chrismc do you mean functionality wise or definition? for node we can define what actually means , what network protocol using, peer nodes discovery using kademlia dht and in future when we will have inclusion of central node vs normal node then we can add more context, about bootstrap nodes and net traversal protocols supported. @erwin1 what do you think?
From cli side we can just define different subcommands to talk to your local node and give information about that

prince-chrismc: Definitely looking for definitions, I'd like to shy away from the technical details here.
Those technical details would go nicely with #462 and they can be linked to from here 👍 

The goal here is to "explaining the folder names to someone randomly clicking around" this should help them navigate to whatever they are more interested in.

We dont want too many locations with details but just enough to help people find them 🤞

src/readme.md
```diff
@@ -1,16 +1,19 @@
 # Pyrsia Library
 
-Principale set of module making up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli)
+Principale set of module which make up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli) components.
```
betarelease, line 3: Principle? Principal?

betarelease (dismissed):

src/readme.md
prince-chrismc: 🤣 

Sorry that's the French spelling

image

prince-chrismc:
```suggestion
Principle set of module which make up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli) components.
```

AbhijithGanesh (dismissed):

src/readme.md
```diff
@@ -1,16 +1,19 @@
 # Pyrsia Library
 
-Principale set of module making up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli)
+Principle set of module which make up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli) components.
```
erwin1, line 3: I'm sorry to come back on this, please ignore if I don't make any sense. I'm not speaking English natively so I usually refrain from these discussions. But 'principle' is a noun and I don't think you can use it like that. So I'd say:

```suggestion
Principal set of modules which make up the [Node](../pyrsia_node) and [CLI](../pyrsia_cli) components.
```
(also plural module)

src/readme.md
erwin1: I actually missed out on this mention here. I think there might be some good p2p and docker candidates to explain here in this section, but I'd need to think about it. You might as well go ahead with this PR and then I'll do another pass later on

src/readme.md
prince-chrismc: I am glad I am not alone in that respect! From googling I found

> When used as an adjective, principal means “main” or “primary,” such as the principal finding in a study. On the other hand, principle is a noun that means a rule, tenet, or basic truth, such as the principle of gravity.

which is definitely what I was originally trying to say

MitaliBo (approved):

PR #430, opened 2022-02-28
Breakdown development workflow from PR template

prince-chrismc:
The goal of this PR is to reduce the burden on contributors looking to get involved.
But it's extremely important to capture the details of developing the project locally.

- Trimmed the PR template + flipped order
- new "dev workflow"
   - old template + missing steps to install rust tooling
- reworked "good pr" so be more human friendly

PR #447, opened 2022-03-02
Fix login name of dependabot to match correctly

prince-chrismc:
Apparently the user login is this... https://api.github.com/repos/pyrsia/pyrsia/pulls/441
It's treated specially in the UI i guess

Template does not make sense! #430 will help 🙏


sbtaylor15 (approved):
/lgtm

chb0github (approved):

PR #449, opened 2022-03-02
Disable project board automation jobs on forks

prince-chrismc:
I got reports of random failure... this is what we did for the other workflows so I applied it here as well


sbtaylor15 (approved):
/lgtm - nice catch

tiainen (approved):

PR #464, opened 2022-03-05
Introduce the `KeyBox` for consensus engine

prince-chrismc:
## PR Checklist

- [x] I've built the code `cargo build`. For major changes, `cargo build --workspace --release` is recommended.

## Description

Towards pyrsia/pyrsia#322

This PR add the `KeyBox` this is the main trait the Aleph needs to verify the signatories of blocks during consensus

https://docs.rs/aleph-bft/latest/aleph_bft/trait.KeyBox.html

> A typical implementation of `KeyBox` would be a collection of `N` public keys, an index `i` and a single private key corresponding to the public key number `i`

### DIFF
--- a/src/blockchain/src/identities/authority_pen.rs
+++ b/src/blockchain/src/identities/authority_pen.rs
@@ -14,25 +14,27 @@
    limitations under the License.
 */
 
-// TODO(prince-chrismc): Re-introduce `NodeIndex` to associate with `PeerId` when adding `KeyBox`
-// use aleph_bft::NodeIndex;
-use libp2p::core::identity::ed25519::Keypair;
+use aleph_bft::NodeIndex;
+use libp2p::core::identity::ed25519::{Keypair, PublicKey};
 
 use super::signature::Signature;
 
 #[derive(Clone)]
 pub struct AuthorityPen {
-    // index: NodeIndex,
+    index: NodeIndex,
     keypair: Keypair,
 }
 
 impl AuthorityPen {
-    pub fn new(/*index: NodeIndex,*/ keypair: Keypair) -> Self {
-        Self {
-            /*index,*/ keypair,
-        }
+    pub fn new(index: NodeIndex, keypair: Keypair) -> Self {
+        Self { index, keypair }
     }
-
+    pub fn index(&self) -> NodeIndex {
+        self.index
+    }
+    pub fn public(&self) -> PublicKey {
+        self.keypair.public()
+    }
     pub fn sign(&self, msg: &[u8]) -> Signature {
         Signature::new(msg, &self.keypair)
     }
@@ -45,7 +47,7 @@ mod tests {
     #[test]
     fn test_auth_pen_sign() {
         let keypair = Keypair::generate();
-        let auth_pen = AuthorityPen::new(keypair.clone());
+        let auth_pen = AuthorityPen::new(0.into(), keypair.clone());
         let signed = auth_pen.sign(b"hello world!");
 
         assert!(keypair.public().verify(b"hello world!", &signed.to_bytes()));
--- a/src/blockchain/src/identities/key_box.rs
+++ b/src/blockchain/src/identities/key_box.rs
@@ -1,0 +1,111 @@
+/*
+   Copyright 2021 JFrog Ltd
+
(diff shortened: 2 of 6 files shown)


src/blockchain/src/identities/authority_pen.rs
```diff
@@ -14,25 +14,27 @@
...
 
 impl AuthorityPen {
-    pub fn new(/*index: NodeIndex,*/ keypair: Keypair) -> Self {
-        Self {
-            /*index,*/ keypair,
-        }
+    pub fn new(index: NodeIndex, keypair: Keypair) -> Self {
+        Self { index, keypair }
+    }
+    pub fn index(&self) -> NodeIndex {
```
chb0github, line 32: So, help me understand something about access - you have a public struct, aren't the fields also public?

chb0github (approved):
I can only assume this works properly.

src/blockchain/src/identities/authority_pen.rs
tiainen: No. From https://doc.rust-lang.org/reference/visibility-and-privacy.html

> By default, everything in Rust is _private_, with two exceptions: Associated items in a `pub` Trait are public by default; Enum variants in a `pub` enum are also public by default.

chb0github (approved):

src/blockchain/src/identities/key_box.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub fn record_authority(&mut self, node_index: NodeIndex, public_key: PublicKey) {
+        self.authority_verifier.save(node_index, public_key);
+    }
+}
+
+#[async_trait]
+impl aleph_bft::KeyBox for KeyBox {
+    type Signature = Signature;
+
+    fn node_count(&self) -> NodeCount {
```
betarelease, line 51: Is this used anywhere? Meaning is there a reason to expose the node_count from within the verifier and making it available?

Also from the authority_verifier code - nodecount is actually referring to not the nodes available but nodes that act as part of the consensus - is that correct? In which case should that be renamed to reflect that purpose. 
(maybe a nitpick)

src/blockchain/src/identities/key_box.rs
```diff
@@ -0,0 +1,111 @@
...
+        self.authority_verifier.node_count()
+    }
+
+    async fn sign(&self, msg: &[u8]) -> Self::Signature {
+        trace!("🖋️ {:?} signing message", self.authority_pen.index());
+        self.authority_pen.sign(msg)
+    }
+
+    fn verify(&self, msg: &[u8], sgn: &Self::Signature, index: NodeIndex) -> bool {
+        trace!(
```
betarelease, line 61: I see that we are using these trace statements in a few places(They are here in place of debug log like statements to help with debugging right?)

And this case now we are exposing the index which I see is being used only for the trace. Do we still need the trace if we have a unit test to test the verify function? For me it added noise when trying to understand this call to verify.

betarelease (approved):
I left a couple of comments. See if you can address them. 

In general though, I like that this was a small and focused PR which I could understand easily. Kudos on keeping it to this size.

src/blockchain/src/identities/key_box.rs
fishseabowl: NodeCount: the total number of known public keys(consensus nodes). 
Ref: https://docs.rs/aleph-bft/latest/aleph_bft/trait.KeyBox.html

fishseabowl (approved):
It looks good for me.

src/blockchain/src/identities/key_box.rs
prince-chrismc: It's not only for tracing, it's being used in the next statement to know "which authoritiy" we are verifying https://github.com/pyrsia/pyrsia/pull/464/files/aaccf0401ad29cc2a78257c3949b586010012bf4#diff-f5ea41848026ac6fc9d0930a7c77d25df2fbff6e0f060fef04e294987dfef1caR66

being as this is part of Aleph trait, it's not possible to return anything that could be more verbose to describe the problem. 

This is an excellent point to revisit when more of the code is migrated!

tiainen (approved):

PR #476, opened 2022-03-09
Write down the proposed goals for the blockchain crate

prince-chrismc:
The goal with this PR is to establish a baseline for what we are working on.
Conceptually what we be producing.


chb0github (changes requested):
Some commentary about meta_data would be good

fishseabowl (commented):
I suggest we point out that only central nodes can submit transactions and submit blocks to the blockchain.

PR #535, opened 2022-03-29
Add support for encoding required to use AlephBFT

prince-chrismc:
## Description

Towards #322 just another step converting the POC/Hackathon before we can add the other half of this puzzle together

This folds in the https://github.com/aleph-zero-foundation/AlephBFT#dependencies requirement of the `paritytech-scale-codec` and this is a foundational step

## PR Checklist

- [x] I've requested a review  from "pyrsia/collaborators"

### Code Contributions

- [x] I've built the code `cargo build --workspace`.
- [x] I've run the unit tests `cargo test --workspace`.

### DIFF
--- a/src/blockchain/Cargo.toml
+++ b/src/blockchain/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-multihash = {version = "0.16.1", features = ["serde-codec", "scale-codec"]}
+multihash = {version = "=0.16.0", features = ["serde-codec", "scale-codec"]}
 primitive-types = "0.11.1"
 bincode = "1.3.3"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/blockchain/src/crypto/hash_algorithm.rs
+++ b/src/blockchain/src/crypto/hash_algorithm.rs
@@ -14,10 +14,11 @@
    limitations under the License.
 */
 
+use codec::{Decode, Encode};
 use multihash::{Code, Multihash, MultihashDigest};
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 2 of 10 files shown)


src/blockchain/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-multihash = {version = "0.16.1", features = ["serde-codec", "scale-codec"]}
+multihash = {version = "=0.16.0", features = ["serde-codec", "scale-codec"]}
```
prince-chrismc, line 7: This needs to be the exact version. because it was bumped passed a major version 

https://github.com/multiformats/rust-multihash/commit/81215bb6427f069fddf6fafb40a126548e1aefa9

This does not play well with other deps that have not moved to the newest version

src/blockchain/src/crypto/hash_algorithm.rs
```diff
@@ -14,10 +14,11 @@
    limitations under the License.
 */
 
+use codec::{Decode, Encode};
 use multihash::{Code, Multihash, MultihashDigest};
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash, Decode, Encode)]
```
fishseabowl, line 21: If possible, can we support only one encoding and decoding format?

fishseabowl (approved):
LGTM 😄

src/blockchain/src/crypto/hash_algorithm.rs
prince-chrismc: We talked about this in the office, there are never mixed (in our dependencies) so there should not be any problems.
However human error is always possible, we can write bad code but that's always true 😆 

TL;DR
- SCALE codec - for consensus
`serde_json` - for everything else

This is get revised when we know more

mseabornIBM (approved):
With my limited knowledge of the blockchain code it seems reasonable.

Issue #505, opened 2022-03-21
Persist the generated libp2p keypair


tiainen:
At the moment, a private/public ed25519 keypair is generated that is used as the PeerId within the libp2p network. However, the PeerId is regenerated each time the pyrsia node is started.

Investigate if there is a good reason to persist the generated keypair and reload it during consecutive launches of the pyrsia node.

betarelease:
Blockchain needs the PeerID.

tiainen:
When reading a bit about peer identity ([Peer Identity](https://docs.libp2p.io/concepts/peer-id/) and [Addressing](https://docs.libp2p.io/concepts/addressing/)), it seems useful to have a persistent PeerId that can be passed in when dialing:

> Now [not only do] my friends know where to find me, anyone they give that address to can verify that the machine on the other side is really me, or at least, that they control the private key for my PeerId.

PR #539, opened 2022-03-30
persist the libp2p keypair

tiainen:
## Description

Fixes pyrsia/pyrsia#505

Stores the keypair that is used for the p2p network in the pyrsia artifact path.

## PR Checklist

- [x] I've requested a review  from "pyrsia/collaborators"

### Code Contributions

- [x] I've built the code `cargo build --workspace`.
- [x] I've run the unit tests `cargo test --workspace`.


src/network/p2p.rs
```diff
@@ -48,13 +49,13 @@ use std::iter;
 ///  * [`EventLoop`]: used to process events received from the p2p [`Swarm`] and commands from the p2p [`Client`]
 pub async fn create_components(
 ) -> Result<(Client, impl Stream<Item = Event>, EventLoop), Box<dyn Error>> {
-    let local_keys = identity::Keypair::generate_ed25519();
+    let local_keypair = keypair_util::create_ed25519();
```
prince-chrismc, line 52: "create" does not really capture the magic this function is doing, maybe this?

```suggestion
    let local_keypair = keypair_util::initiate_ed25519();
```

prince-chrismc (commented):
One tiny suggestions but otherwise LGTM ✅

src/network/p2p.rs
tiainen: I had indeed some trouble to find a good name. This is what the function does: load a key from file, if it's not there, create a new one and save that one for future calls. `initiate` makes it sound like it creates a new keypair each time it's called. Would `load_or_create_ed25519` be any better?

johanvos: I agree create is misleading and initiate is vague, so +1 from me for load_or_create

johanvos (approved):

src/util.rs
```diff
@@ -15,3 +15,4 @@
 */
 
 pub mod env_util;
+pub mod keypair_util;
```
betarelease, line 18: my only suggestion would be to not add stuff to utils. In fact we should figure out how to move env_util into its own concern.

src/util/keypair_util.rs
```diff
@@ -0,0 +1,125 @@
...
+        KEYPAIR_FILENAME
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn load_non_existing_keypair_fails() {
```
betarelease, line 85: good test coverage. Thanks for the diligence in covering these cases.

betarelease (approved):
Once you address my comments feel free to merge. Ping me if you need another review and I can quickly verify.

src/util.rs
tiainen: Could you create an issue for this, including the rationale for why you believe having a util section is bad practice?

Issue #526, opened 2022-03-25
Implement functions to save and load libp2p Keypair from local files 


fishseabowl:
This issue is part of #505. This issue only uses libp2p Keypair encoding and decoding methods to save and load libp2p Keypair. This issue should be enhanced to meet industry security standards/compliance.

PR #545, opened 2022-03-30
Store and load libp2p keypair locally on Blockchain

fishseabowl:
## Description

Fixes pyrsia/pyrsia#526

This PR does use libp2p Keypair encode and decode methods to save and load ed25519 keypair to ~/.block_keypair. Also, on Unix or Mac, set this file permission to 0o600. 

## PR Checklist

### Code Contributions


MitaliBo (dismissed):

src/blockchain/examples/simple_node.rs
```diff
@@ -152,3 +154,106 @@ pub fn write_block(path: &str, block: Block) {
...
+        .write(true)
+        .create(true)
+        .mode(0o600)
+        .open(path)
+        .expect("cannot open file");
+
+    file.write_all(data).expect("write failed");
+}
+
+pub fn read_keypair(path: &String) -> Result<[u8; 64], Box<dyn Error>> {
```
tiainen, line 173: I see that you sometimes match an error and then just return it. For those cases, it might be easier to read by using the [`?` operator](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator).

For instance, the code in this method can be simplified to:

    let mut file = std::fs::File::open(path)?;
    let mut buf = [0u8; 64];
    let n = file.read(&mut buf)?;
    if n == 64 {
        Ok(buf)
    } else {
        Err(Box::new(io::Error::from(io::ErrorKind::InvalidData)))
    }

src/blockchain/examples/simple_node.rs
```diff
@@ -152,3 +154,106 @@ pub fn write_block(path: &str, block: Block) {
         .expect("write failed");
     file.write_all(b"\n").expect("write failed");
 }
+
+pub fn write_keypair(path: &String, data: &[u8; 64]) {
+    use std::fs;
```
tiainen, line 159: In this file, there are three different ways that other paths are brought into scope:

 * using `use` at the top
 * using `use` inside a function
 * specifying the full path

Wouldn't it be better to be consistent and for instance keep all use declarations at the top?

src/blockchain/examples/simple_node.rs
mseabornIBM: You hardcoded the file name in the test code. Won't that break if you change the constant BLOCK_KEYPAIR_FILENAME?

fishseabowl: Good idea. We will enhance it #561

src/blockchain/examples/simple_node.rs
fishseabowl: Thanks for your comment. I will refactor it.

fishseabowl:
> See my comments about hard coded test vaules.

Thanks for your comment. I will refactor it.

tiainen (approved):

MitaliBo (approved):

Issue #557, opened 2022-04-05
apt-get installer fails in docker build


sbtaylor15:
### Steps taken/Steps to reproduce

See [docker build log](https://github.com/pyrsia/pyrsia/runs/5811477736?check_suite_focus=true)
Also, @AbhijithGanesh has further details if needed.

### Expected Behavior

Pyrsia installed in docker image using apt-get and .deb

### Experienced Instance of the bug

Docker CI part of the build

fishseabowl:
repo error?
#6 15.09 Err:2 https://pyrsia.io/repo focal/main amd64 pyrsia amd64 0.1.0+1256
#6 15.09   404  Not Found [IP: 185.199.109.153 443]

sbtaylor15:
@fishseabowl the gh pages site is down for some reason or in the process of being updated.  Moving the repo off GH to GCS will fix this long term. #562

Issue #542, opened 2022-03-30
Tests are being compiled twice in CI


prince-chrismc:
Post-op from #451 we are compiling in one test and running in the other but it sees cargo thinks it needs to be compiled again 😕 
There some investigation here.

![image](https://user-images.githubusercontent.com/16867443/160928530-5dbad940-0a2d-4ebf-b328-bb73b0830341.png)

prince-chrismc:
There are compilation errors being hit in the run step https://github.com/pyrsia/pyrsia/runs/5760954999?check_suite_focus=true#step:9:1072

prince-chrismc:
This seems to be fixed with #564

Checked a few PRs like https://github.com/pyrsia/pyrsia/runs/6157584859?check_suite_focus=true

PR #564, opened 2022-04-05
check to systemctl exists and use cached files for tests

sbtaylor15:
## Description

Fixes pyrsia/pyrsia#557 pyrsia/pyrsia#542

Docker build was failing due to the systemctl not existing in docker.  Added logic to check for systectl on the path before executing it.  Also, added necessary flags to both the `cargo build` and `cargo test` to have them in sync.  They need to be in sync for the caching to work.  Finally, removed the `brew install openssl` for OS/X since it is not needed.

## PR Checklist

- [X] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/good_pr.md)
- [X] I've requested a review  from `pyrsia/collaborators`.

### Code Contributions

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -16,10 +16,14 @@ set -e
 
 # We need to split out build commands into separate calls in order to run them in parallel
 # therefore we cannot use --all-targets since that is equivalent to --lib --bins --tests --benches --examples
-# which disables the parallelizm we are trying achive for speed.
+# which disables the parallelizm we are trying achive for speed. 
+#
+# "cargo build --workspace"  is equivalent to cargo build --workspace --lib --bins
+# adding --lib --bins for clarity
+#
 
-$( cargo build --workspace --verbose --release; echo $? > /tmp/ws.rc ) &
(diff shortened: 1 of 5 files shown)


AbhijithGanesh (approved):

.github/workflows/build.sh
```diff
@@ -23,7 +23,7 @@ set -e
 #
 
 $( cargo build --workspace --lib --bins --verbose --release; echo $? > /tmp/ws.rc ) &
-$( cargo build --workspace --tests --verbose --release 1>/tmp/tests.log 2>&1; echo $? > /tmp/tests.rc ) &
+$( cargo build --workspace --tests --benches --examples --verbose --release 1>/tmp/tests.log 2>&1; echo $? > /tmp/tests.rc ) &
```
AbhijithGanesh, line 26: I am assuming benches and examples pick up the cached files.

sbtaylor15: Those are different targets that need to be built so they are cached.

mseabornIBM (approved):

Issue #577, opened 2022-04-10
Limit coverage job to run only if rust code has changed


VitaliyaIoffe:
Under #479 issue and #565 were limited linting jobs. 
Should also limit runs of the coverage workflow.

PR #589, opened 2022-04-14
Avoid running coverage job if rs files weren't changed 

VitaliyaIoffe:
## Description

This patch allows to avoid run of coverage job if rust files weren't change
For avoiding duplication between workflows I have placed the job separately, which is aimed to provide info about any_changed rust files.

**I'd like to test it more, so, now it is only draft.** 
Examples: 
https://github.com/VitaliyaIoffe/pyrsia/actions/runs/2168406643
https://github.com/VitaliyaIoffe/pyrsia/actions/runs/2168401487
Close pyrsia/pyrsia#577

# PR Checklist

- [ ] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/good_pr.md)
- [ ] I've requested a review  from `pyrsia/collaborators`.

## Code Contributions

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,22 +9,7 @@ permissions:
 jobs:
 
   rust-files:
-      runs-on: ubuntu-latest
-      outputs:
-        any_changed: ${{ steps.changed-rust-files.outputs.any_changed }}
-      steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - name: Get changed rust files
-        id: changed-rust-files
-        uses: tj-actions/changed-files@v18.7
-        with:
-          files: |
-            **/*.rs
-            Cargo.toml
-            Cargo.lock
+    uses: ./.github/workflows/any_changed_rust_files.yml
 
   security-audit:
     needs: rust-files
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -117,9 +117,14 @@ jobs:
         private_key: ${{ secrets.GPG_PRIVATE }}
         key_passphrase: ${{ secrets.GPG_SECRET }}
         github_repository: pyrsia/pyrsia.github.io
-  
+
+  rust-files:
+    uses: ./.github/workflows/any_changed_rust_files.yml
+
   # Coverage wipes the cache in order to instrument the code.  No need to wait for other jobs to run it.
   coverage:
+    needs: rust-files
+    if: needs.rust-files.outputs.any_changed == 'true'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3


.github/workflows/any_changed_rust_files.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  workflow_call:
+    outputs:
+      any_changed:
+        description: "Any changed rust files"
+        value: ${{ jobs.rust-files.outputs.any_changed }}
+
+jobs:
+  rust-files:
+      runs-on: ubuntu-latest
```
sbtaylor15, line 12: @VitaliyaIoffe is the runs-on needed?  If that could be removed then this works for both linux and osx making it 100% generic.

VitaliyaIoffe: Why do we need this workflow for osx? You can call this job from another osx workflow to check that there are changed rust files. 
I think we can provide os type by input value (and it is possible to use default value). 
I'm not sure but as I remember, also, we can do something like this
`runs-on: [runner1, runner2, runner3]`

prince-chrismc: To run multiple os... https://github.community/t/create-matrix-with-multiple-os-and-env-for-each-one/16895/2 :wink:

.github/workflows/any_changed_rust_files.yml
```diff
@@ -0,0 +1,27 @@
+name: Changed rust files
+
+on:
+  workflow_call:
```
prince-chrismc, line 4: Just out of curiosity did you consider a local composite action instead? Usually if it's only one job that's more efficient

https://dev.to/jameswallis/using-github-composite-actions-to-make-your-workflows-smaller-and-more-reusable-476l

.github/workflows/any_changed_rust_files.yml
sbtaylor15: @VitaliyaIoffe just thinking ahead on being able to re-use this `any_changed_rust_files.yaml` in other workflows.  No specific use case in mind at this time.   Leaving it as-is for this PR works for me.

sbtaylor15 (approved):
changes, as they stand, look good.  Minor requested changes can be done later if needed.

VitaliyaIoffe:
> changes, as they stand, look good. Minor requested changes can be done later if needed.

Sorry, it was a busy week. I haven't answered the @prince-chrismc question yet.

.github/workflows/any_changed_rust_files.yml
VitaliyaIoffe: I haven't seen any articles about how many resources are needed for composite vs reusable. Maybe a simple composite action is really better for this change.

prince-chrismc:
I'd prefer to see this as a Composite Action since that is better, but if you don't have the time we can merge this in for now and improve on it later 👍

VitaliyaIoffe:
> I'd prefer to see this as a Composite Action since that is better, but if you don't have the time we can merge this in for now and improve on it later 👍

I am going to change reusable wf to composite under this pr :)

Issue #540, opened 2022-03-30
Store docker manifests in Pyrsia network


erwin1:
Since #363 docker images are requested from the p2p network if they are not locally available.
The same should also happen for the docker manifests that a docker client requests during a 'docker pull' 

These manifest are currently still handled as originally specified:
* retrieve locally
* retrieve from Docker Hub if not available locally

tiainen:
I have split this issue up into some smaller parts:

- #622 
- #623 
- #624 
- #625

PR #615, opened 2022-04-25
Provision manifests over p2p

tiainen:
## Description

Fixes pyrsia/pyrsia#540

This PR changes the sources for retrieving a docker manifest:

1. retrieve from the local artifact manager
2. retrieve from the pyrsia network
3. retrieve from Docker Hub

## PR Checklist

### Code Contributions

### DIFF
--- a/pyrsia_node/src/network/handlers.rs
+++ b/pyrsia_node/src/network/handlers.rs
@@ -37,7 +37,7 @@ pub async fn provide_artifacts(mut p2p_client: Client) {
             p2p_client
                 .provide(&format!(
                     "{}/{}/{}",
-                    package_version.pkg_type, package_version.name, package_version.version
+                    package_version.namespace_id, package_version.name, package_version.version
                 ))
                 .await;
         }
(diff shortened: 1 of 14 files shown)


pyrsia_node/src/network/handlers.rs
```diff
@@ -30,6 +31,18 @@ pub async fn dial_other_peer(mut p2p_client: Client, to_dial: &Multiaddr) {
...
+        for package_version in package_versions.iter() {
+            p2p_client
+                .provide(&format!(
+                    "{}/{}/{}",
+                    package_version.namespace_id, package_version.name, package_version.version
+                ))
+                .await;
+        }
+    }
+
```
fishseabowl, line 45: Is the manifest part of an artifact? Why are they in the same function?

pyrsia_node/src/network/handlers.rs
```diff
@@ -45,12 +58,63 @@ pub async fn handle_request_artifact(
...
+                "This node does not provide artifact {}. Error: {:?}",
+                hash, e
+            ),
+        };
+    }
+
+    None
+}
+
+fn handle_request_blob(hash: &str) -> Option<Vec<u8>> {
```
fishseabowl, line 108: Could you share some documents or examples about manifests and artifacts if possible?

src/docker/v2/handlers/blobs.rs
```diff
@@ -152,21 +145,24 @@ async fn get_blob_from_network(
...
-) -> Result<bool, Rejection> {
+) -> Result<(), RegistryError> {
     let providers = p2p_client.list_providers(hash).await;
     debug!(
         "Step 2: Does {:?} exist in the Pyrsia network? Providers: {:?}",
         hash, providers
     );
-    Ok(match providers.iter().next() {
+
+    match providers.iter().next() {
```
fishseabowl, line 155: Does this mean that the blob (the content of the artifact) is only stored locally on the peer node?

src/node_manager/model.rs
```diff
@@ -20,3 +20,6 @@ pub mod namespace;
 pub mod package;
 pub mod package_type;
 pub mod package_version;
+
+// TODO This will eventually be defined in namespace metadata, after namespace metadata is implemented
+pub const DOCKER_NAMESPACE_ID: &str = "4658011310974e1bb5c46fd4df7e78b9";
```
fishseabowl, line 25: I'm just curious what this is.

pyrsia_node/src/network/handlers.rs
tiainen: Yes, it would be better to keep the distinction between metadata (aka docker container manifests) and the artifacts (aka docker container blobs). I've pinned the naming too much to the docker side instead of using the pyrsia terminology.

src/docker/v2/handlers/blobs.rs
tiainen: No, ultimately, the blob is always stored locally. This part will download the blob from the first peer node that provides it. When it is retrieved, it will be stored locally on this peer as well.

src/node_manager/model.rs
tiainen: This was moved from here: https://github.com/pyrsia/pyrsia/blob/8836c4ad9eaf52b8d1e1b27d416ccdbb694209ea/src/docker/v2/handlers/manifests.rs#L310

Every type of artifact will have it's own namespace. At the moment, we only support docker, so there hasn't been done much implementing around namespaces yet. Which is why there is this hardcoded value for now.

tiainen:
I will split this PR into separate smaller ones.

PR #691, opened 2022-05-12
added summary of the MVP architecture discussions

erwin1:
## Description

This is a summary of the meetings and discussions we had in the last weeks about the MVP milestone.

Preview: https://github.com/erwin1/pyrsia/blob/MVP-arch-docs/docs/developers/pyrsia-high-level-overview.md
But note that the architecture diagram image is broken until this gets merged. (the link has to be absolute for it to work on the website)

- [x] I've read the [contributing guidelines](https://github.com/pyrsia/.github/blob/main/contributing.md).
- [x] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/good_pr.md)
- [x] I've included a good title and brief description along with how to review them.
- [x] I've linked any associated an [issue](https://github.com/pyrsia/pyrsia/issues).
- [x] I've requested a review  from `pyrsia/collaborators`.

PR #692, opened 2022-05-12
Lint markdown documentation and validate links

### LINKED ISSUE #580: Add linting job for markdown files
Looking over #541 I noticed a handful of issues that would have been caught by https://github.com/markdownlint/markdownlint
I think we are at the point that we should step quality checks here as well.

Similar to #479 and #577  this job should only run if `.md` files are changed

***

Quickly looking I found https://github.com/actionshub/markdownlint

and I also noticed https://github.com/gaurav-nelson/github-action-markdown-link-check which might be fun to toss in

prince-chrismc:
## Description

Fixes #580 

This PR adds markdown linting to the docs repo. These job are limited. It also checks to make sure the links actually work (there's a lot of failures right now)
I also tweaked the rust linting to limited to only rust files changed.

These are not required checks, there are a ton of issues and I hope we can fix them moving forward. The goal is not to block PRs just flag issues (which we can adjust if we dont like the rules)

### Screenshots

For the markdown linter, it gives annotations

image

for the links the output is a bit too messy so I skipped that one (though it's more important)

## PR Checklist

### DIFF
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -12,6 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: sudo gem install mdl
+      - run: echo "::add-matcher::.github/workflows/md-docs-matchers.json"
       - run: mdl docs
 
   markdown-link-check:
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - run: gem install mdl
+      - run: sudo gem install mdl
       - run: mdl docs/
 
   markdown-link-check:
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -18,7 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: gaurav-nelson/github-action-markdown-link-check@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/doc-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1 # Basic .md files
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1 # Also check extended format
+        with:
+          file-extension: .mdx
+
+  inclusive-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: run woke
```
erwin1, line 31: we might be running the exact same woke command twice if both .md and .rs files are changed?
It could be placed in its own workflow?

prince-chrismc: Excellent question! You are correct It would be ran twice... I am not sure on what files it works with the docs are a little weak for this tool.

From https://github.com/get-woke/woke

> woke is a text file analysis tool that finds places within your source code that contain non-inclusive language and suggests replacing them with more inclusive alternatives.

I am not sure it even checks Markdown look at https://github.com/get-woke/woke/search?q=markdown

I think making a separate issue might be worthwhile, maybe @sbtaylor15 has more insight?

erwin1 (approved):

.github/workflows/doc-lint.yml
prince-chrismc: Thank you very much!

prince-chrismc:
> which we can adjust if we dont like the rules

There was a good call out

https://github.com/pyrsia/pyrsia/pull/691#issuecomment-1126266596

AbhijithGanesh (approved):
LGTM

PR #771, opened 2022-06-08
fix table and math rendering

sbtaylor15:
## Description

Fix table rendering and use an external resource to render the math equation since the latest version of Katex does not render it at all.  

Known issue with Docusaurus (tried workaround - it does not render the math equation and does not throw any errors):

```
The latest versions of rehype-katex (starting from v6.0.0) has moved to ES Modules, a new module system of JavaScript, which Docusaurus doesn't officially support yet. However, it is possible to import rehype-katex dynamically, using an async config creator. Docusaurus will call this creator function and wait for it to return the config object.
```

## Screenshots (optional)
image

## PR Checklist

### Code Contributions


fishseabowl (approved):
LGTM

mseabornIBM:
yes that function looks right

mseabornIBM (approved):

PR #776, opened 2022-06-13
Definition of Transparent Log Signature Structure

fishseabowl:
## Description

This PR just defines the transparent log signature structure

## PR Checklist

### Code Contributions


src/transparency_log/log.rs
```diff
@@ -29,19 +29,31 @@ enum TransparencyLogError {
...
     operation: Operation,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct SignatureEnvelope {
+    /// The data that is integrity protected
+    payload: Payload,
+    /// The time at which the signature was generated. This is a part of signed attributes
+    signing_timestamp: u64,
+    /// The digital signatres computed on payload and signed attributes
```
johanvos, line 51: signature

johanvos (approved):
I think this is a good start. I have no idea if that structure is the ultimate one that we should use in the end, but I also have no reason why this is not correct, so it looks good.
If it turns out we need more info in that envelope, or that we want to change part of the structure, we can always revisit this.

src/transparency_log/log.rs
fishseabowl: fixed. thanks

mseabornIBM (dismissed):

sbtaylor15 (approved):
/lgtm

sbtaylor15 (approved):
/lgtm

tiainen (approved):

Issue #793, opened 2022-06-21
Transparency log stores transaction logs in a database


tiainen:
### Suggestion details

At the moment the transparency log component stores all the transaction logs as files on the local file system and in a in-memory hash map. The transparency log should instead use a database.

For the database implementation, we will use SQLite. The [rusqlite](https://crates.io/crates/rusqlite) crate can be used to interact with the database.

The transaction log should be stored in a database table with the following columns:

- namespace_specific_id
- artifact_hash
- timestamp
- operation

PR #799, opened 2022-06-23
Use database for transparency log transaction logs

jperedadnr:
## Description

Fixes pyrsia/pyrsia #793 

So far, the transparency log stores all the transaction logs as files on the local file system and in a in-memory hash map. 
With this PR, the transparency log uses now a database instead, which is created on the local system.

For the database implementation, we use SQLite and the [rusqlite](https://crates.io/crates/rusqlite) crate interacts with the database.

The database `payload.db3` has a `payload` table with the following columns:
```
id
hash
timestamp
operation
```

## PR Checklist

- [x ] I've linked any associated an [issue](https://github.com/pyrsia/pyrsia/issues).

### Code Contributions


src/transparency_log/log.rs
```diff
@@ -90,67 +90,127 @@ impl TransparencyLog {
...
+                timestamp: row.get(2)?,
+                operation: {
+                    let op: String = row.get(3)?;
+                    Operation::from_str(&op).unwrap()
+                },
+            })
+        })?;
+
+        if let Some(record) = payload_records.next() {
+            return Ok(record.unwrap());
```
tiainen, line 206: Errors aren't ignored in all other cases, so why do it here? Perhaps rewrite the check on `payload_records` as follows:

    match payload_records.next() {
        Some(Ok(record)) => Ok(record),
        _ => Err(TransparencyLogError::NotFound { id: id.to_string() }.into()),
    }

src/transparency_log/log.rs
```diff
@@ -90,67 +90,127 @@ impl TransparencyLog {
...
         }
+    }
 
-        anyhow::bail!("No payload found with specified ID");
+    fn open_db(&self) -> anyhow::Result<Connection> {
+        fs::create_dir_all(&self.storage_path)?;
+        let payload_storage_path = self.storage_path.to_str().unwrap();
+        let conn = Connection::open(payload_storage_path.to_owned() + "/payload.db3")?;
+        match conn.execute(
+            "CREATE TABLE IF NOT EXISTS payload (
```
tiainen, line 133: I see that an attempt to create the table is being made upon each call to `open_db`. For now this is ok, but we will need to look for a mechanism to automatically update the database when it is needed. This will probably have to be done outside of the pyrsia codebase.

src/transparency_log/log.rs
```diff
@@ -90,67 +90,127 @@ impl TransparencyLog {
...
+                anyhow::bail!("No payload found with specified ID")
+            }
         }
+    }
 
-        anyhow::bail!("No payload found with specified ID");
+    fn open_db(&self) -> anyhow::Result<Connection> {
+        fs::create_dir_all(&self.storage_path)?;
+        let payload_storage_path = self.storage_path.to_str().unwrap();
+        let conn = Connection::open(payload_storage_path.to_owned() + "/payload.db3")?;
```
tiainen, line 131: `transparency_log.db` sounds like a more appropriate name for the database.

jperedadnr: ok

src/transparency_log/log.rs
```diff
@@ -90,67 +90,127 @@ impl TransparencyLog {
...
-            .write(true)
-            .create_new(true)
-            .open(&payload_filename)
-        {
-            Ok(mut payload_file) => {
-                let json_payload = serde_json::to_string(payload)?;
-                payload_file.write_all(json_payload.as_bytes())?;
+        let conn = self.open_db()?;
+
+        let payload_to_db = payload.clone();
```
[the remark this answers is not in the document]
jperedadnr: hmm, it seems it is?

src/transparency_log/log.rs
jperedadnr: ok

tiainen (approved):

erwin1 (approved):

Issue #804, opened 2022-06-24
Implement get_artifact in Transparency Log Service


erwin1:
The transparency log service is the component (mainly used by the Artifact Service) to store and retrieve transparency log information about an artifact.

This issue implements the retrieval of the latest add_artifact transaction for a given artifact, if it exists. 

The method was defined in #802 as

```
pub fn get_artifact(
    &mut self,
    package_type: &PackageType,
    package_type_id: &str,
) -> Result<Transaction, TransparencyLogError> 
```
It follows this procedure:

- search the transaction log database for all transactions with
    - operation: add_artifact or remove_artifact
    - package_type: given as a parameter to this method
    - package_type_id: the PSID also given as a parameter to this method
    - and sort on timestamp descending
- if the latest transaction is remove_artifact or no transaction is found, an error is returned
- otherwise the latest transaction is returned

Issue #803, opened 2022-06-24
Implement add_artifact in Transparency Log Service


erwin1:
The transparency log service is the component (mainly used by the Artifact Service) to store and retrieve transparency log information about an artifact.

This issue implements the `add_artifact` method to request the addition of a 'add_artifact' transaction to the transparency log. 

The method was defined in #802 as

```
pub struct AddArtifactRequest {
    package_type: PackageType,
    package_type_id: String,
    hash: String,
}

pub async fn add_artifact(
        &mut self,
        add_artifact_request: AddArtifactRequest,
        _sender: oneshot::Sender<Result<Transaction, TransparencyLogError>>,
    ) -> Result<(), TransparencyLogError> {
}
}
```

add_artifact will follow this procedure:

- create a `Transaction` including 
    - operation: `add_artifact`
    - package type: provided by the `AddArtifactRequest`
    - package type specific artifact ID (PSID): provided by the `AddArtifactRequest`
    - timestamp: now
    - artifact_hash: provided by the `AddArtifactRequest` 
    - source_hash: provided by the `AddArtifactRequest`
    - artifact_id: Pyrsia specific id of the artifact -> UUID to be generated by the Transparency Log at the time of Transaction creation
    - source_id: Pyrsia specific id of the artifact's source -> UUID to be generated by the Transparency Log at the time of Transaction creation

- request the addition of the transaction in the blockchain by calling `Blockchain.add_block` with the transaction as the payload.

- the transparency log now has to wait for the blockchain component. (it will need to reach consensus among other authorized nodes)

- when consensus is reached, the transparency log service will be notified by the blockchain component, which will
    - store the transaction in the transaction log database
    - notify the caller of add_artifact that the transaction was added successfully

PR #813, opened 2022-06-29
Extend add and get artifact implementations

jperedadnr:
## Description

Fixes pyrsia/pyrsia#803 
Fixes pyrsia/pyrsia#804

This PR adds more fields to the transparency log database: `source_hash`, `artifact_id`, and `source_id`, and renames `hash` to `artifact_hash`.
Both hashes are now added by the ArtifactRequest, and both `add_artifact` and `get_artifact` are updated as described in the issues.

Blockchain call is added as a to-do.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/transparency_log/log.rs
+++ b/src/transparency_log/log.rs
@@ -26,6 +26,7 @@ use std::str::FromStr;
 use std::time::{SystemTime, UNIX_EPOCH};
 use thiserror::Error;
 use tokio::sync::oneshot;
+use uuid::Uuid;
 
 #[derive(Debug, Clone, Error, PartialEq)]
 pub enum TransparencyLogError {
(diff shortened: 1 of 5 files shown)


src/transparency_log/log.rs
```diff
@@ -125,14 +136,20 @@ impl TransparencyLog {
             id: add_artifact_request.package_type_id.to_string(),
```
tiainen, line 136: This should also become a random uuid instead of the package type id.

tiainen: Good point. We should think a bit more about what makes an artifact unique within the entire pyrsia network. We can take this in a separate issue.

tiainen (approved):

erwin1 (approved):

PR #826, opened 2022-07-03
Removing let _  usage from code

### LINKED ISSUE #756: Extra usage of  `let _ =` clutters the code base
There is a lot of places were a variable is declared but the result is ignored. 

There's a few places that's suppressing errors but other are adding noise to the code.

It would be ideal to clean up the usage.

--- 

If you're not going to use the result, then you don't even need `let`

_Originally posted by @chb0github in https://github.com/pyrsia/pyrsia/pull/712#discussion_r887277828_

vss96:
## Description

Fixes #756

This PR removes extra usages of `let _`.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/artifact_service/service.rs
+++ b/src/artifact_service/service.rs
@@ -279,15 +279,15 @@ mod tests {
                     Some(Command::ListProviders { artifact_type: _artifact_type, artifact_hash: _artifact_hash, sender }) => {
                         let mut set = HashSet::new();
                         set.insert(local_peer_id);
-                        let _ = sender.send(set);
+                        sender.send(set);
                     },
                     Some(Command::RequestIdleMetric { peer: _peer, sender }) => {
-                        let _ = sender.send(Ok(PeerMetrics {
+                        sender.send(Ok(PeerMetrics {
                             idle_metric: (0.1_f64).to_le_bytes()
                         }));
                     },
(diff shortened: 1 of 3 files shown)


tiainen:
Removing the assignments is not the solution to fix this issue. Have you read https://github.com/pyrsia/pyrsia/issues/756#issuecomment-1145652326 about why these statements were used? The only solution I can think of, would be to evaluate each statement and see whether we want to:

* keep ignoring the `Result` and hence let the `let _` remain in place
* stop ignoring the `Result` and print a info/warn/error message in case of `Err`
* stop ignoring the `Result` and panic with a message in case of `Err`

betarelease (changes requested):
We don't necessarily want to remove let _ but do what is correct(handle the error, raise an error, leave a log message- based on the situation) for each such case. 

Removing the let _ is poor practice because rust expects the return value to be handle - especially when it could result in an error.

betarelease:
@dasmanas has started looking into this issue as well - https://github.com/pyrsia/pyrsia/issues/756.

It would be a good idea for you to pair with @dasmanas and work on this together @vss96

betarelease:
We discussed this at standup today and there is more context we can share with you  @vss96. 

If you are available we do have daily standup meetings at 11:05AM PST which you can join. You can join the googlegroup - https://groups.google.com/g/pyrsia to receive all the meeting invites and join them.

betarelease:
Closing this since we are working on this in https://github.com/pyrsia/pyrsia/pull/937

PR #835, opened 2022-07-06
make keypair name and location configurable

### LINKED ISSUE #831: p2p_keypair.ser should not be in the artifact location
The p2p_keypair.ser key is being persisted in the artifact directory.  This causes a conflict when trying to provide a key that lives outside of the artifact directory.  The location and name of the key should be configurable.

sbtaylor15:
## Description

Make the location of the key pair configurable so it is not in the artifact directory.

Fixes #831 

This PR makes the key pair configurable by using PYRSIA_KEYPAIR env var. This env var is the full file name for the key pair.  A new key pair will be generated if one does not exist.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/network/p2p.rs
+++ b/src/network/p2p.rs
@@ -14,7 +14,6 @@
    limitations under the License.
 */
 
-use crate::artifact_service::storage::ARTIFACTS_DIR;
 use crate::network::artifact_protocol::{ArtifactExchangeCodec, ArtifactExchangeProtocol};
 use crate::network::behaviour::PyrsiaNetworkBehaviour;
 use crate::network::blockchain_protocol::{BlockUpdateExchangeCodec, BlockUpdateExchangeProtocol};
@@ -22,6 +21,7 @@ use crate::network::client::Client;
 use crate::network::event_loop::{PyrsiaEvent, PyrsiaEventLoop};
 use crate::network::idle_metric_protocol::{IdleMetricExchangeCodec, IdleMetricExchangeProtocol};
 use crate::util::keypair_util;
+use crate::util::keypair_util::KEYPAIR_FILENAME;
 
 use libp2p::core;
 use libp2p::dns;
@@ -38,7 +38,6 @@ use libp2p::yamux;
 use libp2p::Transport;
 use std::error::Error;
 use std::iter;
-use std::path::PathBuf;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tokio_stream::Stream;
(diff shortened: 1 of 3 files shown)


src/util/keypair_util.rs
```diff
@@ -14,25 +14,73 @@
...
+use log::{error, warn};
 use std::error;
 use std::fs;
 use std::io::{self, Read, Write};
-use std::path::{Path, PathBuf};
+use std::panic::UnwindSafe;
+use std::path::Path;
+
+lazy_static! {
+    pub static ref KEYPAIR_FILENAME: String = {
```
tiainen, line 29: Instead of the lazy static variable, can't we load the path from the environment variable in the `load_or_generate_ed25519()` function instead? Apart from the creation of the parent directory, I don't see a real use case for lazy static initialization. And you already took care of that part in the `save_ed25519()` function anyway.

efrisch (approved):
lgtm. 👍

efrisch (changes requested):
Oops. Merge needed first.

src/util/keypair_util.rs
sbtaylor15: We cannot load the variable from inside of `load_or_generate_ed25519()` since we pass in different filenames/paths as part of the test cases parameters to `load_or_generate_ed25519()`.   The code is easier to understand using the lazy_static.

tiainen: Looks like we disagree on readability there. We can refactor the usage of `lazy_static` in another issue if needed.

src/network/p2p.rs
```diff
@@ -110,8 +109,7 @@ use tokio_stream::Stream;
 pub fn setup_libp2p_swarm(
     max_provided_keys: usize,
 ) -> Result<(Client, impl Stream<Item = PyrsiaEvent>, PyrsiaEventLoop), Box<dyn Error>> {
-    let local_keypair =
-        keypair_util::load_or_generate_ed25519(PathBuf::from(ARTIFACTS_DIR.as_str()));
+    let local_keypair = keypair_util::load_or_generate_ed25519(KEYPAIR_FILENAME.as_str());
```
tiainen, line 112: You missed a change that was there in an earlier commit. Unfortunately, those changes got removed because of the forced push. Fortunately, I remembered: you'll need to update the use of `load_or_generate_ed25519` in `pyrsia_node/src/main.rs` (https://github.com/pyrsia/pyrsia/blob/main/pyrsia_node/src/main.rs#L133) as well to use KEYPAIR_FILENAME instead of ARTIFACTS_DIR.

sbtaylor15: @tiainen thanks for catching that.  Got introduced when I tried to remove the lasy_static.  All fixed now.

tiainen (approved):

efrisch (approved):
👍

Issue #711, opened 2022-05-19
AutoNAT protocol implementation


MitaliBo:
AutoNAT protocol helps to define if a node is behind the NAT or not. As of now we are assuming that every node is behind the NAT and so they are connected with relay first and attempt to hole punch for NAT Traversal

betarelease:
Is this the same as your PR for autoNAT? @efrisch

efrisch:
> Is this the same as your PR for autoNAT? @efrisch

@betarelease Yes. https://github.com/pyrsia/pyrsia/pull/848

betarelease:
Also is this the same as #584? Can we close one of them?

Issue #584, opened 2022-04-13
NAT discovery for node with autonat protocol


MitaliBo:
Before starting process for hole punching, the node need to know if it is behind NAT or not and for that we need to have autonat protocol.
We will be using pyrsia central node as boot_node and a server as well for this.

efrisch:
Moved into implementation phase. In fact, we are in code review. Close.

PR #848, opened 2022-07-12
AutoNAT Protocol Implementation

efrisch:
## Description

Fixes pyrsia/pyrsia#711

This PR adds AutoNAT support by integrating with [libp2p-autonat](https://crates.io/crates/libp2p-autonat) and it can be reviewed; you can also test the changes by connecting to gcp-sandbox

## PR Checklist

### Code Contributions


fishseabowl (approved):
LGTM

src/network/p2p.rs
```diff
@@ -112,7 +108,23 @@ pub fn setup_libp2p_swarm(
...
+        .with(libp2p::multiaddr::Protocol::Ip4(
+            std::net::Ipv4Addr::UNSPECIFIED,
+        ))
+        .with(libp2p::multiaddr::Protocol::Tcp(
+            config.p2p_port.parse::<u16>().unwrap_or(44120),
+        ));
+
+    swarm.listen_on(address.clone())?;
+
+    swarm
```
tiainen, line 124: I propose to change this call to be initiated via the p2p client, as is currently done for instance for [listening](https://github.com/pyrsia/pyrsia/blob/e0f36b04e2f8333f72b00e657f301a249684399a/src/network/client.rs#L77) and [dialing](https://github.com/pyrsia/pyrsia/blob/e0f36b04e2f8333f72b00e657f301a249684399a/src/network/client.rs#L91). It could be called `probe` or `add_probe_address` and take the PeerId and Multiaddr to add for probing as a parameter.

```
pub async fn add_probe_address(&mut self, peer_id: &PeerId, probe_addr: &Multiaddr) -> anyhow::Result<()> {
    self.sender
        .send(Command::AddProbe {
            peer_id: *peer_id,
            probe_addr: probe_addr.clone(),
        })
        .await
}
```

And then handle the extra command in the event handler (src/network/event_loop.rs):
```
Command::AddProbe { peer_id, probe_addr } => {
    self.swarm
        .behaviour_mut()
        .auto_nat
        .add_server(peer_id, Some(probe_addr));
```

src/network/p2p.rs
```diff
@@ -112,7 +108,23 @@ pub fn setup_libp2p_swarm(
     let local_keypair =
         keypair_util::load_or_generate_ed25519(PathBuf::from(ARTIFACTS_DIR.as_str()));
 
-    let (swarm, local_peer_id) = create_swarm(local_keypair, max_provided_keys)?;
+    let (mut swarm, local_peer_id) = create_swarm(local_keypair, max_provided_keys)?;
+    let config: CliConfig = get_config().unwrap();
```
tiainen, line 112: This panics when `p2p_port` is not defined in the local pyrsia config. I would remove it from the CliConfig and instead pass it as an argument to the pyrsia node. Akin to the `--peer` argument to dial another pyrsia node, I would add support for an optional argument `--probe <address>` that, when provided will add the address to the autonat behavior. See my comment about that a bit below.

tiainen (changes requested):

src/network/event_loop.rs
```diff
@@ -328,6 +328,16 @@ impl PyrsiaEventLoop {
     async fn handle_command(&mut self, command: Command) {
         trace!("Handle Command: {}", command);
         match command {
+            Command::AddProbe {
+                peer_id,
+                probe_addr,
+                sender: _,
```
tiainen, line 334: You'll need to call `sender.send()` after the server has been added. Otherwise the receiver side in the network client will wait indefinitely. As mentioned in the other comment, we can just send the unit type `()` to notify the receiver that the command was successfully handled: `sender.send(())`.

sbtaylor15 (dismissed):
/lgtm

pyrsia_node/src/main.rs
```diff
@@ -121,8 +121,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
 async fn setup_p2p(mut p2p_client: Client, args: &PyrsiaNodeArgs) -> Result<()> {
     p2p_client.listen(&args.listen_address).await?;
-
-    if let Some(to_dial) = &args.peer {
+    if let Some(to_probe) = &args.probe {
+        info!("Invoking probe");
+        handlers::probe_other_peer(p2p_client.clone(), to_probe).await
+    } else if let Some(to_dial) = &args.peer {
+        info!("Invoking dial");
```
fishseabowl, line 128: What happens if both probe addr and peer addr exist in the CLI? Just ignore the peer addr?

src/blockchain/src/network.rs
```diff
@@ -30,23 +30,18 @@ use futures::{
...
-    mplex,
-    noise,
+    mplex, noise,
     request_response::{
         ProtocolSupport, RequestResponse, RequestResponseCodec, RequestResponseConfig,
         RequestResponseEvent, RequestResponseMessage,
     },
     swarm::{NetworkBehaviourEventProcess, SwarmBuilder},
-    // `TokioTcpTransport` is available through the `tcp-tokio` feature.
-    tcp::{GenTcpConfig, TokioTcpTransport},
```
fishseabowl, line 45: It's better to keep lines 44-45 as a reminder to add `tcp-tokio` feature

src/network/event_loop.rs
efrisch: @tiainen This is ready for re-review.

pyrsia_node/src/main.rs
efrisch: Yes. Peer should go away long term. Or probe should replace it. Long term. Note the next pr will make it optional, and default to "boot.pyrsia.link"

betarelease:
@efrisch Is this a PR for #584 ? Can we mark it as such?

sbtaylor15 (approved):
/lgtm

fishseabowl (approved):

johanvos (approved):

Issue #856, opened 2022-07-14
Blockchain: Refactoring blockchain methods to support fetching and adding payloads


fishseabowl:
The blockchain methods need to be refactored to support other services (e.g. transparency log service) adding and fetching payloads.

PR #857, opened 2022-07-14
Refactoring blockchain methods to support fetching and adding payloads

fishseabowl:
## Description

Fixes pyrsia/pyrsia#856

This PR refactors the following blockchain methods to support fetchting and adding payloads:
```rust
 pub fn last_block(&self) -> Option<Block>
 pub fn add_block(&mut self, payload: Vec<u8>, local_key: identity::Keypair) -> anyhow::Result<()>
 pub fn fetch_payload(&self) -> Vec<Vec<u8>>
 fn commit_block(&mut self, block: Block)
```

## PR Checklist

### Code Contributions

### DIFF
--- a/src/blockchain/src/blockchain.rs
+++ b/src/blockchain/src/blockchain.rs
@@ -15,6 +15,8 @@
 */
 
 use libp2p::identity;
+use libp2p::identity::Keypair::Ed25519;
+use log::debug;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt::{self, Debug, Formatter};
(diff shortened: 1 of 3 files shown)


src/blockchain/src/blockchain.rs
```diff
@@ -105,7 +116,42 @@ impl Blockchain {
         self
     }
 
-    #[warn(dead_code)]
+    pub fn add_new_block(&mut self, payload: Vec<u8>, local_key: identity::Keypair) {
```
tiainen, line 119: Since there is also a method called `add_block`, it would make sense to add some documentation that describes the difference between the two?

src/blockchain/src/blockchain.rs
```diff
@@ -105,7 +116,42 @@ impl Blockchain {
...
+        let submitter = Address::from(local_key.public());
+        let ed25519_key = match local_key {
+            Ed25519(some) => some,
+            _ => {
+                debug!(
```
tiainen, line 124: The caller of this method could potentially be interested in the fact whether the addition succeeded or not? In other words, does it make sense to change the return type to a Result instead?

fishseabowl: Add Result return type. Thanks

erwin1 (approved):

tiainen (approved):

Issue #843, opened 2022-07-11
Upgrade method tcp::TokioTcpConfig to tcp::TokioTcpTransport


fishseabowl:
Upgrade method tcp::TokioTcpConfig to tcp::TokioTcpTransport according to libp2p 0.46.1 release changelog.
Ref: https://docs.rs/libp2p/latest/libp2p/tcp/index.html

PR #858, opened 2022-07-14
Upgrade method tcp::TokioTcpConfig to tcp::TokioTcpTransport

fishseabowl:
## Description

Fixes pyrsia/pyrsia#843

This PR changes the method tcp::TokioTcpConfig to tcp::TokioTcpTransport to upgrade to libp2p 0.46.1

## PR Checklist

### Code Contributions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ hyper-tls = "0.5.0"
 if-watch = "1.1.1"
 itertools = "0.10.3"
 lazy_static = "1.4.0"
-libp2p = { version = "0.45.1", features=["dns-tokio", "tcp-tokio"]}
+libp2p = { version = "0.46.1", features=["dns-tokio", "tcp-tokio"]}
 log = { version = "0.4.17", features = ["max_level_trace", "release_max_level_trace"] }
 maplit = "1.0.2"
 multihash = {version = "0.16.0", features = ["serde-codec"]}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -604,7 +604,7 @@ version = "3.2.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "759bf187376e1afa7b85b959e6a664a3e7a95203415dba952ad19139e798f902"
 dependencies = [
- "heck 0.4.0",
+ "heck",
  "proc-macro-error",
  "proc-macro2",
  "quote",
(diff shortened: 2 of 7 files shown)


tiainen (approved):

Cargo.toml
```diff
@@ -45,7 +45,7 @@ hyper-tls = "0.5.0"
 if-watch = "1.1.1"
 itertools = "0.10.3"
 lazy_static = "1.4.0"
-libp2p = { version = "0.45.1", features=["dns-tokio", "tcp-tokio"]}
+libp2p = { version = "0.46.1", features=["dns-tokio", "tcp-tokio"]}
```
efrisch, line 48: Let's stick with `libp2p = { version = "0.45.1", features=["dns-tokio", "tcp-tokio"]}` for now. Because this breaks libp2p-autonat 0.5.0 (which is still the current version).

efrisch (changes requested):
Hold for now.

Cargo.toml
tiainen: Can you explain how it breaks autonat? From the [libp2p dependencies](https://crates.io/crates/libp2p/0.46.1/dependencies), I can see that `libp2p:0.46.1` depends on `libp2p-autonat:0.5.0` already? I also see that the NAT draft PR [manually includes a dependency](https://github.com/pyrsia/pyrsia/pull/848/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R49) to `libp2p-autonat:0.5.0`, which should not be needed, because the feature is [included by default](https://docs.rs/crate/libp2p/latest/features)?

efrisch: @tiainen  @sbtaylor15 When I ran it with libp2p:0.46.1 it broke the build on my branch. Revisiting now, because the build is broken on my branch (again).

tiainen: The update does indeed break the build. That's the reason for this PR, to fix the breaking changes that were introduced with libp2p:0.46.1. When we merge this PR, you can simply merge these changes into your branch and your code should build without problems. Then you can also remove the hard dependency on libp2p-autonat 0.5.0, because libp2p 0.46.1 already depends on that version.

efrisch: @tiainen  @sbtaylor15  I merged these changes into my branch last week, and it triggered build errors relating to libp2p_core traits not being compatible with libp2p traits of the same names. I will try it again today. Perhaps I did something else incorrectly with that merge.

efrisch: > I've tried it on your branch earlier today and managed to build it.

`cargo build` works. `cargo build --workspace --all-targets` fails. With errors below.

efrisch (approved):
Joeri helped me determine what I was doing wrong. This merges cleanly with the autonat changes on the correct branch. 👍

Issue #886, opened 2022-07-25
Auto-discover when connecting with peer nodes


tiainen:
The rust implementation of the libp2p Kademlia protocol does not discover peers automatically when two nodes are communicating with each other (for instance by using a dial). This is explained in a bit more detail here: https://docs.rs/libp2p-kad/0.38.0/libp2p_kad/index.html#important-discrepancies.

Issue #865 was created from the misconception that two peers would always be connected bi-directionally when peer A dials peer B. The fact is that in this case, peer A will contain peer B in its list of connected peers, but not the other way around.

A second, unrelated, misconception, is that the connection is closed after some time. That is also perfect valid behavior, because there is no point in keeping connections alive, when no actual communication is happening between two peers. The default keepalive timeout is set to 10 seconds, but can be configured if needed. The peer will still be listed in the routing table, so that whenever communication is needed later on, a new connection will be established between the two nodes to allow them to communicate with each other.

To resolve the issue of auto-discovery of peers, we need to introduce the libp2p `Identify` protocol again. We already had an implementation of this before, but that was removed in #790 when fixing issue #663.

PR #887, opened 2022-07-25
add identify protocol to discover listener address of other peers

tiainen:
## Description

Fixes pyrsia/pyrsia#886

This PR re-introduces the libp2p Identify protocol. The protocol is used as a peer discovery mechanism for the Kademlia protocol by updating discovered addresses in Kademlia's internal routing tables.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/network/event_loop.rs
+++ b/src/network/event_loop.rs
@@ -21,6 +21,7 @@ use crate::network::client::command::Command;
 use crate::network::idle_metric_protocol::{IdleMetricRequest, IdleMetricResponse, PeerMetrics};
 use libp2p::core::PeerId;
 use libp2p::futures::StreamExt;
+use libp2p::identify::IdentifyEvent;
 use libp2p::kad::{GetClosestPeersOk, GetProvidersOk, KademliaEvent, QueryId, QueryResult};
 use libp2p::multiaddr::Protocol;
 use libp2p::request_response::{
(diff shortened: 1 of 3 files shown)


betarelease (approved):
This was super fast turnaround in getting this fixed. I like that the solution looks simple and easy to understand.

src/network/event_loop.rs
```diff
@@ -500,6 +523,10 @@ mod tests {
             .boxed();
 
         let behaviour = PyrsiaNetworkBehaviour {
+            identify: identify::Identify::new(identify::IdentifyConfig::new(
+                "ipfs/1.0.0".to_owned(),
```
fishseabowl, line 527: Can ipfs be changed to another name?

fishseabowl (approved):
LGTM

src/network/event_loop.rs
tiainen: I am not sure if the name actually matters, but it resembles the value of the core libp2p identify protocol: https://docs.libp2p.io/concepts/protocols/#core-libp2p-protocols

Also, the docs suggested this value, that's why I ended up using it: https://docs.rs/libp2p/latest/libp2p/identify/struct.IdentifyConfig.html

I suppose that by choosing apyrsia specific name, that only other pyrsia nodes will be identified and not other ipfs nodes? Do you have a suggestion for the protocol name? I am thinking of `pyrsia/id/1.0.0` which matches closely with the id of the libp2p identify. protocol.

fishseabowl: pyrsia/id/1.0.0 is fine. We can also refactor this name later if it breaks your PR. thanks

Issue #818, opened 2022-06-29
Verification service: implement validation of add_artifact transaction


erwin1:
When a `verify(payload)` is called with a transaction payload (= transparency log) of operation `add_artifact` the verification service will:

- based on the package type (from the transparency log)
- start a build using the build service and wait for the result
- when the build finishes, hash the binary and source artifacts and compare them to the hashes in the transparency log
  - simple hashing might not be sufficient for non-reproducible builds
  - if the hashes match, return success to the caller of this function

PR #896, opened 2022-07-27
Verify build

tiainen:
## Description

Fixes pyrsia/pyrsia#818

This PR implements the `verify_transaction` function on the `VerificationService`. The implementation only works for builds that are reproducible, where the hashes of the built artifacts will always be the same.

It also introduces a `BuildEventClient` that is used by the different services to send build events on the build event loop.

## PR Checklist

### Code Contributions


betarelease:
@efrisch and @fishseabowl added you as reviewers too

betarelease (approved):
The only thing that was not clear to me was how is verify_build being used and I found that the arch document delegates that to the consensus algorithm.

fishseabowl (approved):
LGTM

src/verification_service/service.rs
```diff
@@ -14,66 +14,733 @@
...
     /// Verify a build for the specified transaction. This method is
     /// used to be able to reach consensus about a transaction that
     /// is a candidate to be committed to the blockchain.
     pub async fn verify_transaction(
-        &self,
-        _transaction: Transaction,
-        _sender: oneshot::Sender<Result<VerificationResult, VerificationError>>,
-    ) -> Result<(), VerificationError> {
+        &mut self,
+        transaction: &Transaction,
```
fishseabowl, line 102: I guess the verify service doesn't care about the transaction, only the payload of the transaction. So it's better to get the payload: &Vec<u8> as a parameter.

mseabornIBM:
Could I get a link to the architecture document that explains all this?

fishseabowl:
> Could I get a link to the architecture document that explains all this?

@mseabornIBM You can start by looking at https://github.com/pyrsia/pyrsia/blob/main/docs/developers/pyrsia-architecture-and-use-cases.md#verification-service

src/verification_service/service.rs
tiainen: That will probably be enough, if the verification service only ever needs to handle payloads that are transparency logs.

tiainen: Changed the function signature to accept a slice instead of the blockchain transaction.

fishseabowl (approved):
Thanks for the update

johanvos (approved):

PR #919, opened 2022-08-03
Ignores the advisory for owning_ref

betarelease:
- owning_ref is a dependency of libp2p and has a vulnerability that does not have an upgrade path
- this breaks our audit check and hence we need to ignore it till the upgrade is available.
- this vulnerability has been around since Jan 2022 and also the codebase has not been updated for 2 years
- so our mileage is going to be meh.
- https://github.com/libp2p/rust-libp2p/issues/2794

## Description

Ignores vulnerability that is causing audit to fail on github actions due to issues with owning_ref

This PR does..

Ignores the advisory for owning_ref
- owning_ref is a dependency of libp2p and has a vulnerability that does not have an upgrade path
- this breaks our audit check and hence we need to ignore it till the upgrade is available.
- this vulnerability has been around since Jan 2022 and also the codebase has not been updated for 2 years
- so our mileage is going to be meh.
- https://github.com/libp2p/rust-libp2p/issues/2794

## PR Checklist

### Code Contributions

- Although I have run the precommit script that runs all the above the github action change will only be testable on github

### DIFF
--- a/.github/workflows/.cargo/audit.toml
+++ b/.github/workflows/.cargo/audit.toml
@@ -2,4 +2,4 @@
 # A good example is at https://docs.rs/crate/cargo-audit/0.10.0/source/audit.toml.example
 
 [advisories]
-ignore = [] # advisory IDs to ignore
+ignore = [RUSTSEC-2022-0040] # advisory IDs to ignore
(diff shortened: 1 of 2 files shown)


efrisch (approved):
Looks good to ignore the blocking owner-ref package dependency.

fishseabowl (approved):

johanvos (approved):
Although I typically want to avoid bypassing security warnings, I agree that this is the best option at this moment, while keeping track of the upstream issue in libp2p.

AbhijithGanesh (approved):
LGTM

Issue #938, opened 2022-08-09
Correctly handle docker builds from tag


tiainen:
When the build service is handling a build from source for a docker image based on a tag, it will generate one artifact for the image manifest where the PSAID is equal to the PSID. When a docker pull is initiated for this image, the docker client will in fact try to request the manifest via the manifest's digest as well. Currently, this will result in a 404 because the pyrsia node can't map the digest to an artifact. To solve this, an extra artifact is needed for the manifest, when the docker image is loaded via a tag instead of a digest.

For example, handling the image with PSID `alpine:3.15.2` will result in one artifact:
 - PSID: `alpine:3.15.2`
 - PSAID: `alpine:3.15.2`
 - hash: `73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`

The following extra artifact must be created too:
 - PSID: `alpine:3.15.2`
 - PSAID: `alpine@sha256:73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`
 - hash: `73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`

Note: this extra artifact is not needed if the build from source was requested directly via the image digest. In that case, we can simply keep the single artifact we are already providing.

PR #939, opened 2022-08-09
Create two artifacts for docker image manifest

tiainen:
## Description

Fixes pyrsia/pyrsia#938

This PR updates the build service to create two artifacts for the docker manifest when the build is requested via the docker image tag instead of a digest: one mapping the image tag and a second one mapping the manifest's digest.

I've also taken the opportunity to update the PSAID's for docker manifests and blobs so that they are more alike and always include the name of the docker image.

For example, below is a list of produced artifacts before and after the change in this PR for the docker image with PSID `alpine:3.15.2`.

Before:
 1. PSAID: `alpine:3.15.2`
     hash: `73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`
 2. PSAID: `sha256:9c842ac49a39fe42e71a623183fe7fb7c759d59029e9a8e7851c357c7d8a86f8`
     hash: `9c842ac49a39fe42e71a623183fe7fb7c759d59029e9a8e7851c357c7d8a86f8`
 3. PSAID: `sha256:3aa4d0bbde192bfaba75f2d124d8cf2e6de452ae03e55d54105e46b06eb8127e`
     hash: `3aa4d0bbde192bfaba75f2d124d8cf2e6de452ae03e55d54105e46b06eb8127e`

After:
 1. PSAID: `alpine:3.15.2`
     hash: `73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`
 2. PSAID: `alpine@sha256:73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`
     hash: `73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`
 3. PSAID: `alpine@sha256:9c842ac49a39fe42e71a623183fe7fb7c759d59029e9a8e7851c357c7d8a86f8`
     hash: `9c842ac49a39fe42e71a623183fe7fb7c759d59029e9a8e7851c357c7d8a86f8`
 4. PSAID: `alpine@sha256:3aa4d0bbde192bfaba75f2d124d8cf2e6de452ae03e55d54105e46b06eb8127e`
     hash: `3aa4d0bbde192bfaba75f2d124d8cf2e6de452ae03e55d54105e46b06eb8127e`

## PR Checklist

### Code Contributions

### DIFF
--- a/src/build_service/service.rs
+++ b/src/build_service/service.rs
@@ -183,10 +183,18 @@ impl BuildService {
             let (artifact_location, artifact_hash) = hash_and_store_data(build_path, &artifact)
                 .map_err(|e| BuildError::Failure(build_id.to_owned(), e.to_string()))?;
 
-            let artifact_specific_id = match package_type {
+            let artifact_specific_ids = match package_type {
                 PackageType::Docker => {
                     if artifact_url.ends_with("/manifest") {
-                        package_specific_id.to_owned()
+                        if package_specific_id.contains('@') {
+                            vec![package_specific_id.to_owned()]
+                        } else {
+                            let docker_image_name = get_docker_image_name(&package_specific_id);
+                            vec![
+                                package_specific_id.to_owned(),
+                                format!("{}@sha256:{}", docker_image_name, artifact_hash),
+                            ]
+                        }
                     } else {
                         let artifact_filename = match artifact_url.rfind('/') {
                             Some(position_slash) => {
@@ -194,10 +202,12 @@ impl BuildService {
                             }
                             None => artifact_url,
                         };
-                        match artifact_filename.rfind('.') {
+                        let blob_digest = match artifact_filename.rfind('.') {
                             Some(position_dot) => String::from(&artifact_filename[..position_dot]),
                             None => artifact_filename,
-                        }
+                        };
+                        let docker_image_name = get_docker_image_name(&package_specific_id);
+                        vec![format!("{}@{}", docker_image_name, blob_digest)]
                     }
                 }
                 PackageType::Maven2 => {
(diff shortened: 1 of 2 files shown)


src/build_service/service.rs
```diff
@@ -266,6 +278,17 @@ fn calculate_hash(bytes: &[u8]) -> String {
     hex::encode(sha256.finalize())
 }
 
+fn get_docker_image_name(package_specific_id: &str) -> String {
+    let docker_image_name = match package_specific_id.rfind('@') {
+        Some(position_at) => &package_specific_id[..position_at],
+        None => match package_specific_id.rfind(':') {
```
johanvos, line 284: Do we want to log something here? Not sure it's useful though.

fishseabowl (commented):
Can you provide more details about PSID and PSAID? What's the difference?

tiainen:
It should be documented here: https://github.com/pyrsia/pyrsia/blob/main/docs/developers/pyrsia-architecture-and-use-cases.md#concepts-used-in-this-document.

But it seems that it does not yet mention the difference between the two. I'll quickly explain here.

A PSID is a package specific id. It identifies a single package of a supported ecosystem. Examples are alpine:3.15.2 for docker and commons-codec:commons-codec:1.15 for maven.

A package can result in a multitude of artifacts. Each of those are identified by a PSAID, a package specific artifact id. Examples are the docker manifest, config and list of blobs for a docker image, and the jar, sha1 and md5 checksum files, sources and javadoc jars for a maven artifact.

fishseabowl (approved):
lgtm

johanvos (approved):

src/build_service/service.rs
tiainen: That's indeed a valid remark. This piece of code handles three use cases:

1. the psid contains `@`: the docker image is specified by the manifest digest, e.g. `alpine@sha256:73c155696fe65b68696e6ea24088693546ac468b3e14542f23f0efbde289cc97`
2. the psid contains `:`: the docker image is specified by a tag, e.g. `alpine:3.15.2`
3. the psid does not contain any of the above: the docker image is specified by name only which defaults to the `latest` tag, e.g. `alpine`

There are no issues with supporting digests, since they point to a fixed image. The problem could come with supporting tags, since they are allowed to be moved and are therefore not guaranteed to always point to the same image. And the `latest` tag is actually not supported yet. I'll write this down in more detail in a follow-up issue.

PR #953, opened 2022-08-11
Unexpected peer ID - incorrect address (clashing peer ID warning added to logs)

karolh2000:
## Description

The original idea was to somehow prevent a situation when two pairing nodes have the same peer ID however it turned out that this approach makes little sense and I don't think is necessary. IMHO simple warning on both sides (the dialed node and the pairing node) should be enough to highlight the problem. The PR/change is pretty straightforward and low risk.

Fixes pyrsia/pyrsia#

To warn users about "clashing" peer IDs and why node fails to start (in this scenario).

## PR Checklist

- [ ] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/get_involved/good_pr.md)

### Code Contributions

- [ ] I've made sure my rust toolchain is current `rustup update`.


src/network/event_loop.rs
```diff
@@ -331,6 +331,9 @@ impl PyrsiaEventLoop {
             SwarmEvent::ConnectionClosed { .. } => {}
             SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
                 if let Some(peer_id) = peer_id {
+                    if peer_id == *self.swarm.local_peer_id() {
+                        warn!("The dialed node have the same peer ID as the peer node: '{}', both nodes require unique peer ID.", peer_id);
```
betarelease, line 335: Just an english change but otherwise PR is good

" The dialed node has the same peer ID...."

betarelease (approved):

src/network/event_loop.rs
karolh2000: @betarelease Fixed, thanks for pointing this out.

src/network/event_loop.rs
```diff
@@ -331,6 +331,9 @@ impl PyrsiaEventLoop {
             SwarmEvent::ConnectionClosed { .. } => {}
             SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
                 if let Some(peer_id) = peer_id {
+                    if peer_id == *self.swarm.local_peer_id() {
+                        warn!("The dialed node has the same peer ID as the peer node: '{}', both nodes require unique peer ID.", peer_id);
```
tiainen, line 335: `peer node` does not sound like a correct term. Is `current node` or `local node` perhaps better? `dialer node` also came up in my mind, but that might be confusing when having both "dialed" and "dialer" in the same message.

```suggestion
                        warn!("The dialed node has the same peer ID as the current node: '{}'. Please make sure that every node has a unique peer ID.", peer_id);
```

karolh2000: @tiainen makes sense, updated.

tiainen (approved):

betarelease (approved):

fishseabowl (approved):

PR #1015, opened 2022-08-26
Remove assay

betarelease:
## Description
This PR removes assay library since we do not use that anymore in our code

## PR Checklist

### Code Contributions

### DIFF
--- a/pre-commit.bat
+++ b/pre-commit.bat
@@ -3,7 +3,7 @@ echo Running Pyrsia pre-commit validation.
 echo This might take sometime, please do not interrupt if the screen is blank.
 
 IF %1.==. GOTO NoClean
-IF %1==clean echo Cleaning old build artifacts 
+IF %1==clean echo Cleaning old build artifacts
 IF %1==clean cargo clean
 
 :NoClean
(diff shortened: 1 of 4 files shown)


pre-commit.bat
```diff
@@ -13,10 +13,6 @@ cargo audit
 IF %ERRORLEVEL% NEQ 0 (ECHO Cargo audit failed. &Exit /b 1)
 cargo clippy
 IF %ERRORLEVEL% NEQ 0 (ECHO Cargo clippy failed. &Exit /b 1)
-rustup update
```
fishseabowl, line 16: Could double check that it is in sync with github actions? I'm just worried about it would cause PR checks to fail.

betarelease: I removed `rustup update` which also bumps us to the new versions of rust. We would like to update to new versions of rust only when we decide it is a good time and not on every commit. So this needs to go. In case you do find that we run `rust update` in the actions we should remove that as well. I will take a look there.

sbtaylor15 (approved):
/lgtm

fishseabowl (approved):
lgtm

Issue #679, opened 2022-05-10
Implement block storage on filesystem


erwin1:
The blockchain component itself will need access to stored blocks (e.g. to make them available in the p2p network). Therefore the Blockchain component itself will be responsible for storing blocks. 
Block will be stored on filesystem. 1 block = 1 json file.

The block storage itself is local to the blockchain component. It does not need an interface for use in other components.

(Arrow 11 in -https://pyrsia.io/docs/developers/pyrsia-high-level-overview/)

fishseabowl:
Currently PR#1021 is a slightly different implementation that saves all blockchain data to a file, where each line is a block.

PR #1021, opened 2022-08-29
Save blockchain data to local file system

fishseabowl:
## Description

Fixes pyrsia/pyrsia#679

This PR Implement blockchain storage on filesystem by implementing the chain structure save method.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/blockchain/src/error.rs
+++ b/src/blockchain/src/error.rs
@@ -1,0 +1,33 @@
+/*
+   Copyright 2021 JFrog Ltd
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+use thiserror::Error;
+
+use crate::structures::header::Ordinal;
+
+#[derive(Clone, Debug, Error, Eq, PartialEq)]
+pub enum BlockchainError {
+    #[error("Blockchain Error: {0}")]
+    Error(String),
+    #[error("Invalid blockchain length: {0}")]
+    InvalidBlockchainLength(usize),
+    #[error("Invalid blockchain ordianl data: {0}")]
+    InvalidBlockchainOrdianl(Ordinal),
+    #[error("Failed to load from: {0}")]
+    LoadFailed(String),
+    #[error("Failed to save on : {0}")]
+    SaveFailed(String),
+}
--- a/src/blockchain/src/error.rs
+++ b/src/blockchain/src/error.rs
@@ -23,6 +23,8 @@ use crate::structures::header::Ordinal;
 pub enum BlockchainError {
     #[error("IO Error")]
     IOError(#[from] io::Error),
+    #[error("Blockchain start postion: {0} is greater than end postion: {1} ")]
+    InvalidBlockchainPosition(usize, usize),
     #[error("Invalid blockchain length: {0}")]
     InvalidBlockchainLength(usize),
     #[error("Invalid blockchain Ordinal: {0}")]
--- a/src/blockchain/src/error.rs
+++ b/src/blockchain/src/error.rs
@@ -23,14 +23,8 @@ use crate::structures::header::Ordinal;
 pub enum BlockchainError {
     #[error("IO Error")]
     IOError(#[from] io::Error),
-    #[error("Blockchain Error: {0}")]
-    Error(String),
     #[error("Invalid blockchain length: {0}")]
     InvalidBlockchainLength(usize),
-    #[error("Invalid blockchain ordianl data: {0}")]
-    InvalidBlockchainOrdianl(Ordinal),
-    #[error("Failed to load from: {0}")]
-    LoadFailed(String),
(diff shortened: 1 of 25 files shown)


src/blockchain/src/error.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Debug, Error)]
+pub enum BlockchainError {
+    #[error("IO Error")]
+    IOError(#[from] io::Error),
+    #[error("Blockchain Error: {0}")]
+    Error(String),
+    #[error("Invalid blockchain length: {0}")]
+    InvalidBlockchainLength(usize),
+    #[error("Invalid blockchain ordianl data: {0}")]
+    InvalidBlockchainOrdianl(Ordinal),
```
betarelease, line 31: Misspelled. Should be `InvalidBlockChainOrdinal`

src/blockchain/src/error.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use std::io;
+use thiserror::Error;
+
+use crate::structures::header::Ordinal;
+
+#[derive(Debug, Error)]
+pub enum BlockchainError {
+    #[error("IO Error")]
+    IOError(#[from] io::Error),
```
betarelease, line 25: We have a few error types which are not used in the following code. Keep only the error types that are used in the code below.

src/blockchain/src/structures/chain.rs
```diff
@@ -140,4 +215,18 @@ mod tests {
...
+        let mut chain: Chain = Default::default();
+        let keypair = identity::ed25519::Keypair::generate();
+        let transactions = vec![];
+        let block = Block::new(HashDigest::new(b""), 0, transactions, &keypair);
+        chain.add_block(block.clone());
+        assert_eq!(1, chain.len());
+        assert!(chain.save_block(0, 0, temp_file.to_string()).await.is_ok());
+        assert!(fs::remove_file(temp_file).await.is_ok());
+        Ok(())
+    }
```
betarelease, line 231: Need tests for all the error conditions as well.

src/blockchain/src/structures/chain.rs
```diff
@@ -140,4 +215,18 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_save_last_block() -> Result<(), String> {
+        let temp_file = "./blockchain.json";
```
johanvos, line 221: wouldn't it be safer to store this in a real tmp directory? What if the test fails midway, and the file is not removed?

src/blockchain/src/structures/chain.rs
```diff
@@ -140,4 +215,18 @@ mod tests {
...
+        let mut chain: Chain = Default::default();
+        let keypair = identity::ed25519::Keypair::generate();
+        let transactions = vec![];
+        let block = Block::new(HashDigest::new(b""), 0, transactions, &keypair);
+        chain.add_block(block.clone());
+        assert_eq!(1, chain.len());
+        assert!(chain.save_block(0, 0, temp_file.to_string()).await.is_ok());
```
johanvos, line 228: this is a good test, but it doesn't check the contents of the file. I think it would be great if there were more tests that checked if the file contains the expected content.

fishseabowl: Sure. you're right. I plan to create another PR to implement the load/read methods and then check the contents of the file.

src/blockchain/src/structures/chain.rs
betarelease: Yes this should go to a configurable test/tmp directory

src/blockchain/src/structures/chain.rs
```diff
@@ -44,10 +47,83 @@ impl Chain {
...
+                }
+
+                if end > block.header.ordinal {
+                    warn!("The end ordinal {:?} out of bounds ", end);
+                    end = block.header.ordinal;
+                }
+            }
+        }
+
+        let start_pos = match self.get_block_position(start) {
```
tiainen, line 91: This can be shortened with the `?` operator:

```
let start_pos = self.get_block_position(start)?;
let end_pos = self.get_block_position(end)?;
```

src/blockchain/src/structures/chain.rs
```diff
@@ -44,10 +47,83 @@ impl Chain {
...
+    ) -> Result<(), BlockchainError> {
+        match self.last_block() {
+            None => return Err(BlockchainError::InvalidBlockchainLength(self.len())),
+            Some(block) => {
+                if start > block.header.ordinal {
+                    return Err(BlockchainError::InvalidBlockchainLength(self.len()));
+                }
+
+                if end > block.header.ordinal {
+                    warn!("The end ordinal {:?} out of bounds ", end);
```
tiainen, line 85: Why not just return an error in this case as well?

src/blockchain/src/structures/chain.rs
```diff
@@ -44,10 +47,83 @@ impl Chain {
...
+            .create(true)
+            .write(true)
+            .read(true)
+            .append(true)
+            .open(file_path)
+            .await;
+
+        match file {
+            Ok(mut file) => {
+                file.write_all(&serde_json::to_vec(&self.blocks[start_pos..=end_pos]).unwrap())
```
tiainen, line 111: You probably also want to check that `start_pos` is not negative and that `start_pos` is less than `end_pos`?

fishseabowl: thanks. 1. `start_pos` is usize. 2. When `start_pos` is greater than `end_pos`, an error will be reported. I would add a check whether `start_pos` is less than `end_pos`.

betarelease (approved):

tiainen (approved):

dasmanas (approved):

Issue #1020, opened 2022-08-29
Allow transparency log PackageType `None`


erwin1:
We need to be able to create a transparency log without PackageType in issues #1004 (for milestone 0.2) 

It will also be needed for #955 which is not in milestone 0.2, but this extra `None` type needs to be added now so we don't complicate things.

PR #1026, opened 2022-08-30
Add PackageType None

jperedadnr:
## Description

Fixes pyrsia/pyrsia#1020

This PR adds `PackageType::None`, so transparency logs can have transactions purely related to node operations (like adding an authorized node),  that don't have anything to do with either Docker or Maven artifacts.

To complete the matches on `PackageType`, I've just added a `panic`, since those cases shouldn't be reached. But this could be refactor in some different way if needed.

## PR Checklist

### Code Contributions


erwin1 (approved):

src/artifact_service/model.rs
```diff
@@ -29,6 +29,7 @@ use serde::{Deserialize, Serialize};
     strum_macros::EnumString,
 )]
 pub enum PackageType {
+    None,
     Docker,
     Maven2,
 }
```
AbhijithGanesh, line 35: Can the whole issue be approached from an `Option` perspective? We can unwrap the result later.

src/build_service/mapping/service.rs
```diff
@@ -266,4 +269,18 @@ mod tests {
...
+        let mapping_service = MappingService::new("");
+
+        let package_type = PackageType::None;
+        let package_specific_id = "";
+
+        mapping_service
+            .get_mapping(package_type, package_specific_id)
+            .await
+            .unwrap();
+    }
```
AbhijithGanesh, line 285: Again, from an [`Option`](https://doc.rust-lang.org/std/option/) perspective, its easier to handle

AbhijithGanesh (dismissed):

src/artifact_service/model.rs
jperedadnr: Do you mean that we shouldn't add `PackageType::None`, in the first place, but use `pub package_type: Option<PackageType>` in the `TransparencyLog` struct?

That might work, but it will require some changes. I'll see if that would work.

efrisch: I think `None` is just an unfortunate name. We should call it `Metadata` or similar.

efrisch (approved):
I think this looks fine. Might want to rename `None`.

src/artifact_service/model.rs
fishseabowl: Maybe `AuthorityNode` or similar is better.

jperedadnr: I've just pushed some changes that use `pub package_type: Option<PackageType>`. Node operations have nothing to do with Artifacts and packages, and therefore it makes sense to have `package_type = None` for these. 

This implies having `NULL` values in the database, which "can" be read only via error (Err -> None).

I can't extend the tests as we still don't have API for adding nodes yet.

src/transparency_log/log.rs
```diff
@@ -387,8 +394,13 @@ impl TransparencyLogService {
             Ok(TransparencyLog {
                 id: row.get(0)?,
                 package_type: {
-                    let pt: String = row.get(1)?;
-                    PackageType::from_str(&pt).unwrap()
+                    match row
```
tiainen, line 397: This would turn any error into a None value. You might want to explicitly check for a Null sql type? That way, a different sql error would still correctly bubble up. Something like:

```
package_type: {
    let value: Value = row.get(1)?;
    match value {
        Value::Text(str) => Ok(Some(PackageType::from_str(&str).unwrap())),
        Value::Null => Ok(None),
        _ => Err(rusqlite::Error::InvalidColumnType(
            1,
            "package_type".to_owned(),
            value.data_type(),
        )),
    }?
},
```

jperedadnr: Yes, that makes sense, thanks.

tiainen (approved):

erwin1 (approved):

src/artifact_service/model.rs
jperedadnr: @AbhijithGanesh I've done the changes you requested, can you please review? It seems the PR is blocked otherwise. Thanks!

PR #979, opened 2022-08-17
Autonat bootstrap

efrisch:
## Description

Fixes pyrsia/pyrsia#944

This PR implements a bootstrap process by performing a http get to boot.pyrsia.link/status and then kicking off an autonat probe by default. You can test it by
```sh
curl -s http://boot.pyrsia.link/status | jq ".peer_addrs[0]" | tr -d '"'
DEV_MODE=on RUST_LOG=info ./target/debug/pyrsia_node -H 0.0.0.0 -p 7888 -L /ip4/0.0.0.0/tcp/44000
```

## PR Checklist

- [x] I've read the [contributing guidelines](https://github.com/pyrsia/.github/blob/main/contributing.md).
- [x] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/get_involved/good_pr.md)
- [x] I've included a good title and brief description along with how to review them.
- [x] I've linked any associated an [issue](https://github.com/pyrsia/pyrsia/issues).

### Code Contributions

- [x] I've built the code `cargo build --all-targets` successfully.
- [x] I've run the unit tests `cargo test --workspace` and everything passes.
- [x] I've made sure my rust toolchain is current `rustup update`.

PR #1042, opened 2022-09-06
1006 integrate blockchain

### LINKED ISSUE #1006: Integrate TransparencyLog component with Blockchain
In the `add_artifact` of the TransparencyLog component, as a final step, it now stores the TransparencyLog.

This must be changed to: 
- call the blockchain `add_block` method and wait for the Blockchain component to perform its work
- when the block is committed, the transparency log will be notified by the blockchain, after which the transparency log component stores the transaction. this is covered in another issue

tiainen:
## Description

Fixes #1006

This PR adds a corresponding block to the blockchain when a log with the `add_artifact` operation is added.

## PR Checklist

### Code Contributions

### DIFF
--- a/pyrsia_node/src/main.rs
+++ b/pyrsia_node/src/main.rs
@@ -60,12 +60,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
     debug!("Start p2p event loop");
     tokio::spawn(event_loop.run());
 
-    debug!("Create pyrsia services");
-    let artifact_service =
-        setup_pyrsia_services(blockchain.clone(), p2p_client.clone(), &args).await?;
-
     debug!("Create blockchain service component");
     let blockchain_service = setup_blockchain_service(p2p_client.clone())?;
+
+    debug!("Create pyrsia services");
+    let artifact_service =
+        setup_pyrsia_services(blockchain_service.clone(), p2p_client.clone(), &args).await?;
 
     debug!("Setup HTTP server");
     setup_http(&args, artifact_service.clone());
@@ -205,7 +205,7 @@ fn setup_blockchain_service(p2p_client: Client) -> Result<Arc<Mutex<BlockchainSe
 }
 
 async fn setup_pyrsia_services(
-    blockchain: Arc<Mutex<Blockchain>>,
+    blockchain_service: Arc<Mutex<BlockchainService>>,
     p2p_client: Client,
     args: &PyrsiaNodeArgs,
 ) -> Result<Arc<Mutex<ArtifactService>>> {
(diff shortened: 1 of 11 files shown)


pyrsia_node/src/main.rs
```diff
@@ -164,7 +166,7 @@ async fn load_peer_addrs(peer_url: &str) -> anyhow::Result<String> {
                     }
                     let peer_addrs: String = vec_jv[0].to_string();
```
fishseabowl, line 167: Currently only the first peer_addr in vec_jv is supported?

src/artifact_service/service.rs
```diff
@@ -269,42 +268,93 @@ mod tests {
...
-        let (build_event_sender, _build_event_receiver) = mpsc::channel(1);
+        (command_receiver, p2p_client)
+    }
+
+    fn create_artifact_service<P: AsRef<Path>>(
+        artifact_path: P,
+        p2p_client: Client,
+    ) -> (mpsc::Receiver<BuildEvent>, ArtifactService) {
+        let transparency_log_service =
+            create_transparency_log_service(&artifact_path, p2p_client.clone());
```
fishseabowl, line 308: From what I understand, transparent_log_service and p2p client in your unit test currently use different key pairs. Is my understanding correct?

fishseabowl (approved):
lgmt

betarelease:
This is quite a big PR. Can you do a demo/walkthrough at the meeting tomorrow? I will reserve some time towards the end.

pyrsia_node/src/main.rs
tiainen: That seems to be correct, but that code was introduced already with https://github.com/pyrsia/pyrsia/pull/979. I only fixed two clippy warnings that suggested using Vec's `is_empty` function instead of comparing its size.

src/artifact_service/service.rs
tiainen: Your understanding is correct, but it doesn't have any implications. In fact, I don't think the p2p client actually needs to have the peer id at all.

tiainen:
@betarelease the actual important change is in the `log.rs` file where we now call the blockchain service to add the new block: https://github.com/pyrsia/pyrsia/pull/1042/files/ac14a6fbbe840f95784bff45c2c7ac373003da74#diff-4a1f585b39b8f1f634dddd19b914bbd2ec735ea9173e31b0df868d7ea0b1d7b9R190-R195

The reason so many changes are here, is because the setup for the transparency log service now also requires the blockchain service. This in turn also changed the way an artifact service is created. These change then needs to be applied to all the tests that create an artifact service or transparency log service.

dasmanas (approved):
lgtm

PR #1062, opened 2022-09-16
Minimize usage of mutexes

### LINKED ISSUE #988: Providing a file blocks the Pyrsia node
Since we do AutoNat bootstrapping by default, a locally started pyrsia node is automatically connected to the 'Pyrsia network'. 

As a result, I've found that when connected to the central node(s), providing files (e.g. after a build from source) on the p2p network blocks the Pyrsia node from doing anything else.

Steps to reproduce:

- follow the instructions in one of the build from source tutorials (https://pyrsia.io/docs/tutorials/build_from_source_maven/)
- after the build succeeds and the files are provided on the network, try to run: `pyrsia -s` (or hit http://localhost:7888/status)
- The result is that this last command will hang, waiting for something happening inside Pyrsia.

Alternatively, this can be reproduced by simply starting a node:
```
RUST_LOG=debug DEV_MODE=on PYRSIA_ARTIFACT_PATH=/tmp/pyrsia cargo run --package pyrsia_node
```

and then run curl http://localhost:7888/peers

it then gets stuck in p2p_client.list_peers().await

tiainen:
## Description

Fixes #988

This PR fixes an issue where incoming requests to the pyrsia node could be blocked, waiting to get a mutex lock on the artifact service. It does this by removing the usage of Mutex where possible. The only place where we still use one, is for passing around an instance of the BlockchainService. The other services that don't have any reference to memory can safely be cloned. The rest of the services that do have data are only accessed from a single place in code and don't need to be shared.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/artifact_service/service.rs
+++ b/src/artifact_service/service.rs
@@ -16,7 +16,6 @@
 
 use super::model::PackageType;
 use super::storage::ArtifactStorage;
-use crate::blockchain_service::service::BlockchainService;
 use crate::build_service::error::BuildError;
 use crate::build_service::event::BuildEventClient;
 use crate::build_service::model::BuildResult;
@@ -25,7 +24,6 @@ use crate::transparency_log::log::{
     AddArtifactRequest, TransparencyLog, TransparencyLogError, TransparencyLogService,
 };
 use anyhow::{bail, Context};
-use libp2p::identity::Keypair;
 use libp2p::PeerId;
 use log::info;
 use multihash::Hasher;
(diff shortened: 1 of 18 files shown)


erwin1 (approved):

src/artifact_service/service.rs
```diff
@@ -292,37 +284,36 @@ mod tests {
         0x4f,
     ];
 
-    fn create_p2p_client() -> (mpsc::Receiver<Command>, Client) {
+    fn create_p2p_client(keypair: &Keypair) -> (mpsc::Receiver<Command>, Client) {
```
jperedadnr, line 287: Most of the tests use `create_p2p_client` and `create_artifact_service`.
As a follow-up, it might be possible to refactor this and move them to `test_util`?

jperedadnr (approved):
Looks good

PR #1086, opened 2022-09-23
Only add public addresses into DHT

### LINKED ISSUE #1084: Only add public addresses to the DHT routing table
When starting a node with the default parameters, it will connect with one of the boot nodes. If that node tries to execute a query on the Kademlia DHT, it will take a while before completing or will eventually even time out.

Looking at the swarm events, one can see that this is because the node is trying to dial other nodes that are known in the Kademlia DHT's routing table:

    Handle SwarmEvent: Dialing(PeerId("12D3KooWFU...cxp"))
    Handle SwarmEvent: OutgoingConnectionError { peer_id: Some(PeerId("12D3KooWFU...cxp")), error: Transport([("/ip4/172.25.0.2/tcp/44000/p2p/12D3KooWFU...cxp", Other(Custom { kind: Other, error: Other(A(A(Transport(Os { code: 113, kind: HostUnreachable, message: "No route to host" })))) }))]) }

It is trying to connect with nodes that are not publicly reachable and when a lot of these nodes are present in the routing table, it will take a long time to finish a DHT query.

Currently, we insert any node that connects to the network into the routing table of the DHT. To fix the issue, we should only add addresses of nodes that are publicly reachable by other nodes.

tiainen:
## Description

Fixes #1084

This PR changes the behavior of adding addresses to the Kademlia DHT routing table.

Previously, we waited for an Identify event from an external node and then used the listener addresses from the event and add them to the Kademlia DHT. This gave issues when the addresses are not reachable by the other nodes, causing the Kademlia queries to take a very long time or even time out.

Instead of these Identify events, we now wait for an update in the NAT status as reported by the autonat behavior. If the status changes to NatStatus::Public, the associated public address is added to the Kademlia DHT routing table.

I've also changed the way that a list of peers is retrieved. Instead of querying the Kademlia DHT for a list of known peers, we use the list of nodes that are currently connected to the swarm. Since the autonat behavior is probing the boot node frequently enough, the connection idle timeout will never be reached, hence the node will continously be connected to the boot node.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/network/event_loop.rs
+++ b/src/network/event_loop.rs
@@ -114,12 +114,18 @@ impl PyrsiaEventLoop {
             AutonatEvent::OutboundProbe(..) => {}
             AutonatEvent::StatusChanged { old, new } => {
                 info!("Autonat status changed from {:?} to {:?}", old, new);
-                if let NatStatus::Public(address) = new {
-                    let local_peer_id = *self.swarm.local_peer_id();
-                    self.swarm
-                        .behaviour_mut()
-                        .kademlia
-                        .add_address(&local_peer_id, address);
+                match new {
+                    NatStatus::Public(address) => {
+                        let local_peer_id = *self.swarm.local_peer_id();
+                        self.swarm
+                            .behaviour_mut()
+                            .kademlia
+                            .add_address(&local_peer_id, address);
+                    }
+                    NatStatus::Private => {
+                        // todo: setup relay listen address
+                    }
+                    NatStatus::Unknown => {}
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


sbtaylor15 (approved):
/lgtm

erwin1 (approved):

src/network/event_loop.rs
```diff
@@ -377,6 +370,11 @@ impl PyrsiaEventLoop {
             } => {
                 if endpoint.is_dialer() {
                     if let Some(sender) = self.pending_dial.remove(&peer_id) {
+                        self.swarm
+                            .behaviour_mut()
```
betarelease, line 374: I did not find a test for this in this PR. This should be tested somehow either in unit or integration tests.

Issue #997, opened 2022-08-22
Add ExternalDNS to helm charts


sbtaylor15:
From inside of a pod we need to update the Google Cloud DNS with the external IP address for the node, `pyrsia-node-1.pyrsia.link` since the services are not pinned to a static ip.

PR #1091, opened 2022-09-23
Add External DNS to the helm charts

sbtaylor15:
## Description

Update the helm charts and doc to include External DNS.  The Oracle section is still work-in-progress and will be handled in a separate PR.

Fixes pyrsia/pyrsia#997

## PR Checklist

### Code Contributions

### DIFF
--- a/installers/helm/pyrsia-node/Chart.yaml
+++ b/installers/helm/pyrsia-node/Chart.yaml
@@ -2,8 +2,8 @@
 apiVersion: v2
 description: Pyrsia Decentralized Package Network
 name: pyrsia-node
-appVersion: 0.1.0-2252
-version: 0.1.0-2252
+appVersion: 0.1.0+2252
+version: 0.1.0+2252
 home: https://pyrsia.io
 icon: https://raw.githubusercontent.com/pyrsia/pyrsia.github.io/main/static/img/icon-pyrsia-torch.svg
 keywords: 
--- a/installers/helm/pyrsia-node/Chart.yaml
+++ b/installers/helm/pyrsia-node/Chart.yaml
@@ -2,8 +2,8 @@
 apiVersion: v2
 description: Pyrsia Decentralized Package Network
 name: pyrsia-node
-appVersion: 0.1.0-2236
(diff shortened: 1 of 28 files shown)


installers/helm/pyrsia-node/Chart.yaml
```diff
@@ -2,8 +2,8 @@
 apiVersion: v2
 description: Pyrsia Decentralized Package Network
 name: pyrsia-node
-appVersion: 0.2.0
-version: 0.2.8
+appVersion: 0.1.0-2236
```
betarelease, line 5: The chart version dropped to 0.1. It should go up or stay the same. No?

installers/helm/pyrsia-node/templates/service.yaml
```diff
@@ -3,6 +3,8 @@ kind: Service
 apiVersion: v1
 metadata:
   name: pyrsia-node-0
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: pyrsia-node-0.pyrsia.link
```
betarelease, line 7: maybe in the future this should be injectable from a values.yaml. Because we should not hardcode node_ids.

betarelease (approved):
Other than the couple of comments I have we should move forward with this since we need this to do deployments.

installers/helm/pyrsia-node/Chart.yaml
sbtaylor15: The chart version is now being set based on the GH Actions.  And the GH Actions are using the version from the cargo.toml.  That is why it went backwards.

installers/helm/pyrsia-node/templates/service.yaml
sbtaylor15: Changed the chart to derive the nodes names from the values.yaml.  Defaults to pyrsia-node and can be overridden at deployment time by using --set name=my-pyrsia-node at the helm command line.

betarelease (dismissed):

betarelease (approved):

efrisch (approved):
lgtm. 👍

Issue #1082, opened 2022-09-22
Support for blockchain initialization on Pyrsia


fishseabowl:
Added init-blockchain subcommand to Pyrsia nodes to create a genesis block and initialize the blockchain. Without this subcommand, we assume this Pyrsia node will pull blocks from other Pyrsia nodes.

PR #1111, opened 2022-09-28
Implement blockchain initialization

fishseabowl:
## Description

Fixes pyrsia/pyrsia#1082
Fixes pyrsia/pyrsia#1128

This PR implements blockchain initialization and adds a new command line argument --init_blockchain for the first Pyrsia node to create the genesis block.

## PR Checklist

### Code Contributions


pyrsia_node/src/main.rs
```diff
@@ -215,21 +219,42 @@ async fn load_peer_addrs(peer_url: &str) -> anyhow::Result<String> {
...
+
+        let blockchain_ed25519_keypair = match blockchain_keypair {
+            libp2p::identity::Keypair::Ed25519(v) => v,
+            _ => {
+                bail!("Keypair Format Error");
+            }
+        };
+        // Refactor to overloading(trait) later
+        blockchain_service = BlockchainService::init_first_blockchain_node(
+            &local_ed25519_keypair,
```
tiainen, line 248: Both these keypairs are exactly the same. Is it the intention that the blockchain keypair is different from the keypair of the node?

pyrsia_node/src/main.rs
```diff
@@ -351,17 +376,19 @@ fn setup_http(
...
-    #[test]
-    fn setup_blockchain_success() {
-        let (p2p_client, _, _) = p2p::setup_libp2p_swarm(100).unwrap();
-        let blockchain_service = setup_blockchain_service(p2p_client);
-        assert!(blockchain_service.is_ok());
+async fn pull_block_from_other_nodes(
+    blockchain_service: Arc<Mutex<BlockchainService>>,
+    args: &PyrsiaNodeArgs,
+) -> anyhow::Result<()> {
+    if let Some(other_node_addr) = &args.peer {
```
tiainen, line 383: I believe we should also pull the blocks when we are connecting to a bootstrap node with `--bootstrap-url`?

pyrsia_node/src/main.rs
```diff
@@ -351,17 +376,19 @@ fn setup_http(
...
+        if !args.init_blockchain {
+            debug!("Blockchain start pulling from other nodes");
+            let other_peer_id = libp2p::PeerId::try_from_multiaddr(other_node_addr).unwrap();
+            let mut blockchain_service = blockchain_service.lock().await;
+            blockchain_service
+                .init_pull_from_others(&other_peer_id)
```
tiainen, line 389: I tested this and the blocks are indeed pulled from the other node correctly. However, I suppose we also need to update the transparency log with the information from the pulled blocks?

src/blockchain_service/service.rs
```diff
@@ -148,12 +221,33 @@ impl BlockchainService {
...
     }
+
+    pub async fn init_pull_from_others(
+        &mut self,
+        other_peer_id: &PeerId,
+    ) -> Result<(), BlockchainError> {
+        // Always start with the genesis block
+        let ordinal = self.query_blockchain_ordinal(other_peer_id).await?;
+
+        for block in self
```
tiainen, line 232: While testing this, I lowered the `max_size` parameter in `blockchain_protocol` from 1_500_000 to 1_500 and noticed that it will indeed stop processing after it has read 1_500 bytes. The protocol should be updated to be able to handle a blockchain that grew beyond 1_500_000 bytes in size. This is mostly an issue when reading the response (https://github.com/pyrsia/pyrsia/blob/main/src/network/blockchain_protocol.rs#L73).

pyrsia_node/src/main.rs
sbtaylor15: @tiainen - At this point they are both sharing the same key pair.  We will need to have a second key for the blockchain that is managed from a KMS.

tiainen: Ok, thanks for clarifying this.

sbtaylor15: @tiainen - Added #1121 to track this

pyrsia_node/src/main.rs
fishseabowl: @tiainen Thanks for your comments. How about `to_probe`? I would create another issue for these.
The relevant code is as follows:
```rust
async fn setup_p2p(mut p2p_client: Client, args: &PyrsiaNodeArgs) -> anyhow::Result<()> {
    p2p_client.listen(&args.listen_address).await?;
    if let Some(to_probe) = &args.probe {
        info!("Invoking probe");
        handlers::probe_other_peer(p2p_client.clone(), to_probe).await
    } else if let Some(to_dial) = &args.peer {
        info!("Invoking dial");
        handlers::dial_other_peer(p2p_client.clone(), to_dial).await
    } else if args.listen_only {
        info!("Pyrsia node will listen only. No attempt to connect to other nodes.");
        Ok(())
    } else {
        info!("Looking up bootstrap node");
        let peer_addrs = load_peer_addrs(&args.bootstrap_url).await?;
        // Turbofish! https://doc.rust-lang.org/std/primitive.str.html#method.parse
        let pa = peer_addrs.parse::<libp2p::Multiaddr>()?;
        info!("Probing {:?}", pa);
        handlers::probe_other_peer(p2p_client.clone(), &pa).await
    }
}
```

pyrsia_node/src/main.rs
fishseabowl: Yes. After updating blockchain, I guess transparency log including the list of artifacts and the list of authorized nodes should be updated.

pyrsia_node/src/main.rs
tiainen: Yes, that would be needed as well. In fact, any time you either call `dial` or `probe`, we would want to pull the blocks. Would it make sense that we only start pulling after we successfully established a connection with the peer that is to be dialed/probed? In that case, the pull could then be initiated here instead: https://github.com/pyrsia/pyrsia/blob/main/src/network/event_loop.rs#L437-L448. Although, I don't mind if you keep this in main as it is now, and just call it from these three different code branches. Thoughts?

src/blockchain_service/service.rs
fishseabowl: @tiainen Thanks for your test. I created another issue #1126 for this.

pyrsia_node/src/main.rs
tiainen: I've created #1130 as a follow-up issue to tackle this.

tiainen (approved):

erwin1 (approved):
tested and works fine

Issue #1101, opened 2022-09-27
Try to trigger a build when an artifact was requested but not found


erwin1:
Follow-up for #1004 where the build trigger mechanism was implemented.

Now this build trigger can also be used when an artifact is requested but not found in the transparency log.

Detail description:
* when a docker image is requested
* or when a maven artifact is requested
* but not found in the transparency log
* call the request build method for the docker image/maven artifact

* the request build will do what it needs to do: either starting a build (on an authorized node) or send a build request to an authorized node (on a regular node)

erwin1:
nice to have for 0.2

PR #1158, opened 2022-10-07
Trigger build if artifact is not found in transparency log

jperedadnr:
## Description

Fixes pyrsia/pyrsia#1101

This PR triggers a build in the case that an artifact requested is not found in the transparency log.

To prevent the same artifact is being built twice if two simultaneous request are done (via HEAD and GET for Docker), 
the fetch manifest has been split into two methods, one that checks if the artifact is present or not, but doesn't trigger a build, and one that checks if the artifact is present or not and triggers a build in case it is not.

The build request is async, and therefore the user that request a missing artifact will still get a non found artifact error. Subsequent requests to that artifact will work for this or any other user once the build is done. In case of Docker, the fallback to download from DockerHub is still in place, so the first user won't have a bad experience.

## PR Checklist

### Code Contributions


sbtaylor15 (approved):
/lgtm

src/java/maven2/handlers/maven_artifacts.rs
```diff
@@ -68,15 +95,12 @@ fn get_package_specific_artifact_id(full_path: &str) -> Result<String, anyhow::E
     if pieces.len() < 4 {
         bail!(format!("Error, invalid full path: {}", full_path));
     }
-    let file_name = pieces.pop().unwrap();
-    let version = pieces.pop().unwrap();
-    let artifact_id = pieces.pop().unwrap();
+    let file_name = pieces.pop().unwrap().to_string();
```
fishseabowl, line 98: I'm worried that if something goes wrong, the main process will stop.

src/java/maven2/handlers/maven_artifacts.rs
```diff
@@ -39,11 +46,16 @@ pub async fn handle_get_maven_artifact(
 
     // request artifact
     debug!(
-        "Requesting artifact for id {}",
-        package_specific_artifact_id
+        "Requesting artifact for id {} and artifact id {}. If not found a build will be requested",
```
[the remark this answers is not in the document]
jperedadnr: Right, I'll add a more detailed message.

src/java/maven2/handlers/maven_artifacts.rs
jperedadnr: Okay, I'll add some guards here.

fishseabowl (approved):

efrisch (approved):
👍

PR #1195, opened 2022-10-13
Handling large blockchain data initialization

fishseabowl:
## Description

This PR is a follow-up to PR #1206 to load the entire blockchain data in segments at regular node initialization when the blockchain data is larger than 10MB.

## PR Checklist

### Code Contributions


src/blockchain_service/service.rs
```diff
@@ -168,22 +168,51 @@ impl BlockchainService {
...
+            other_peer_id
+        );
+        let mut cur_start = start;
+
+        while end > cur_start {
+            if end >= cur_start + 4 {
+                let mut buf: Vec<u8> = vec![];
+                buf.push(cmd);
+                buf.append(&mut serialize(&cur_start).unwrap());
+                let cur_end = cur_start + 4;
```
betarelease, line 185: I see magic number 4. Why do we need 4 and how does that help in the case of block sizes? Can we make this a constant with a good name and use that constant so that we know why it is used and can change it appropriately?

src/network/blockchain_protocol.rs
```diff
@@ -51,12 +51,12 @@ impl RequestResponseCodec for BlockchainExchangeCodec {
     where
         T: AsyncRead + Unpin + Send,
     {
-        let buffer = read_length_prefixed(io, 1_500_000).await?;
+        let buffer = read_length_prefixed(io, 30_000_000).await?;
```
betarelease, line 54: Magic number again - can we move this to a correctly named constant and use it so that when we change it we change it at one place. Also if we bump it from 1.5MB to 30MB how will that impact the size and downloads. We certainly need a performance test for this. Please add that to the integration tests repo. Also it would be good to understand why we need to go to 30MB from 1.5MB and what would be the next jump. 1.5MB sounds too low if the next jump was 30MB

src/blockchain_service/service.rs
fishseabowl: Good comments. For a balance of performance and network bandwidth, I think 5 blocks in one message is a good choice. We should also adjust this number in performance tests to find better numbers. Also, we limit the block size to 1M, which is proven by Bitcoin

src/network/blockchain_protocol.rs
fishseabowl: Thanks for your comments. 1.5MB and 30MB are just message size limits to avoid flooding the network with large messages. It is also within the range of yamux default setting max_buffer_size of 1MB per stream and default max_num_streams of 8192. Also, if a block size is limited to 1MB, and a blockchain message contains a maximum of 5 blocks, then 30MB is sufficient.

betarelease:
Also for this my question is - does this happen when we try to bulk load images to authority node? What if the builds are triggered one by one using the pyrsia cli - will we see the same issue?

fishseabowl:
> Also for this my question is - does this happen when we try to bulk load images to authority node? What if the builds are triggered one by one using the pyrsia cli - will we see the same issue?

No, it's only triggered when a regular node starts up for the first time, while an authorized node has a large blockchain data.

betarelease:
I think this PR can be closed now? @fishseabowl

fishseabowl:
> I think this PR can be closed now? @fishseabowl

@betarelease This PR is a follow-up solution. It will download up to 10MB of blockchain data at a time until all data is downloaded.

src/blockchain_service/service.rs
```diff
@@ -182,27 +187,62 @@ impl BlockchainService {
         end: Ordinal,
     ) -> Result<Vec<Block>, BlockchainError> {
         let cmd = BlockchainCommand::PullFromPeer as u8;
+        let mut blocks = Vec::new();
+        log::debug!(
```
tiainen, line 191: This debug statement has been duplicated.

src/blockchain_service/service.rs
```diff
@@ -182,27 +187,62 @@ impl BlockchainService {
...
+            "Blockchain pull blocks {:?} to {:?} from PeerID {:?}",
+            start,
+            end,
+            other_peer_id
+        );
 
-        let mut buf: Vec<u8> = vec![];
+        let start_mutex = Arc::new(Mutex::new(start));
```
tiainen, line 198: What is the reason for using an Arc with a Mutex? Looks like you can just use a plain integer type?

src/blockchain_service/service.rs
```diff
@@ -389,6 +429,32 @@ mod tests {
         test_util::tests::teardown(tmp_dir);
     }
 
+    #[tokio::test(flavor = "multi_thread")]
```
tiainen, line 432: This test doesn't test the `pull_block_from_other_nodes` method, but instead just tests `pull_blocks`?

auto_build_pull_20.sh
```diff
@@ -0,0 +1,5 @@
+for i in {1..20}
```
tiainen, line 1: Not sure if this script should be committed?

src/blockchain_service/service.rs
fishseabowl: This is to ensure that each block is added sequentially across multiple threads

tiainen: But you are wrapping **local** variables in an `Arc<Mutex>`, so threading is not an issue here?

betarelease:
Can we close this for now since we are working on refactoring some of this @fishseabowl ?

fishseabowl:
Temporarily close this PR to focus on refactoring

PR #1173, opened 2022-10-10
Pyrsia MacOS Installer using Homebrew Automation

dasmanas:
## Description

Fixes pyrsia/pyrsia#1165

This PR takes care the automation of Homebrew installer archive generation and upload to cloud bucket. It supports nightly and stable version releases.

The directory structure Google Cloud Bucket `gs://homebrewrepo` looks like below.
```
.
├── nightly
│   ├── x86_64
│   ├── arm64
├── stable
│   ├── x86_64
│   ├── arm64

```

## PR Checklist

- [x] I've read the [contributing guidelines](https://github.com/pyrsia/.github/blob/main/contributing.md).
- [x] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/get_involved/good_pr.md)
- [x] I've included a good title and brief description along with how to review them.
- [x] I've linked any associated an [issue](https://github.com/pyrsia/pyrsia/issues).

### Code Contributions

- [x] I've built the code `cargo build --all-targets` successfully.
- [x] I've run the unit tests `cargo test --workspace` and everything passes.
- [x] I've made sure my rust toolchain is current `rustup update`.

PR #1208, opened 2022-10-17
Revert "Pyrsia MacOS Installer using Homebrew Automation"

dasmanas:
Reverts pyrsia/pyrsia#1173


dasmanas:
Will be handled through https://github.com/pyrsia/pyrsia/pull/1210

Issue #1190, opened 2022-10-13
Create an option to show config + file location


dasmanas:
### Suggestion details

`pyrsia config -s` or `pyrsia config --show` should display config file location along with the configuration.

betarelease:
Instead `pyrsia config -s` also shows the path of the config. So that we do not need to use another command for showing specific details.

PR #1245, opened 2022-10-26
configuration file path information to terminal

dasmanas:
## Description

Fixes pyrsia/pyrsia#1190

This PR implements the functionality to show the location of the config file along with the content of the config file. You can test using `pyrsia config --show` and it should show something like following.
```shell
Config file path: /Users/manasd/Library/Preferences/rs.pyrsia-cli/default-config.toml
host = '0.0.0.0'
port = '7888'
disk_allocated = '10 GB'
```
Information to `Config file path:` has been added in this PR.

## PR Checklist

### Code Contributions

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -288,11 +288,15 @@ jobs:
       with:
         toolchain: stable
         override: true
+      run: |
+        cargo clean
+        cargo build --all-targets --workspace
+      shell: /bin/bash -e {0}
 
     - name: Run cargo-tarpaulin
       uses: actions-rs/tarpaulin@v0.1
       with:
-        args: '--all-targets --workspace -- --test-threads 1'
+        args: '--skip-clean --all-targets --workspace -- --test-threads 1'
 
     - name: Upload to codecov.io
       uses: codecov/codecov-action@v3
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -288,6 +288,10 @@ jobs:
       with:
         toolchain: stable
         override: true
+
+    # This cargo build creates all binaries under target/debug/ directory. These files get used by unit test cases
+    # designed to test CLI commands. This step is to complement cargo tarpaulin which doesn't create binaries.
+    - name: Cargo build for Cargo Tarpaulin
       run: |
         cargo clean
         cargo build --all-targets --workspace
--- a/pyrsia_cli/src/cli/handlers.rs
+++ b/pyrsia_cli/src/cli/handlers.rs
@@ -53,6 +53,17 @@ pub fn config_add() {
 }
 
 pub fn config_show() {
+    match config::get_config_file_path() {
+        Ok(path_buf) => {
+            println!(
+                "Config file path: {}",
+                path_buf.into_os_string().into_string().unwrap()
+            )
+        }
+        Err(error) => {
+            println!("Error retrieving config file path:       {}", error);
+        }
+    }
     let result = config::get_config();
     match result {
         Ok(config) => {
--- a/pyrsia_cli/src/cli/handlers.rs
+++ b/pyrsia_cli/src/cli/handlers.rs
@@ -14,6 +14,7 @@
    limitations under the License.
 */
 
+use crate::CONF_FILE_PATH_MSG_STARTER;
 use lazy_static::lazy_static;
 use pyrsia::cli_commands::config;
 use pyrsia::cli_commands::node;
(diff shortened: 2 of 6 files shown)


dasmanas:
Test case failing while trying to run pyrsia CLI from `pyrsia/target/debug/pyrsia` using Command::cargo_bin.

```
test tests::test_config_show_file_path_info_availability ... FAILED

 failures:
---- tests::test_config_show_file_path_info_availability stdout ----
 thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CargoError { cause: Some(NotFoundError { path: "/pyrsia/target/debug/pyrsia" }) }', pyrsia_cli/src/main.rs:85:52
```

dasmanas:
> Test case failing while trying to run pyrsia CLI from `pyrsia/target/debug/pyrsia` using Command::cargo_bin.
> 
> ```
> test tests::test_config_show_file_path_info_availability ... FAILED
> 
>  failures:
> ---- tests::test_config_show_file_path_info_availability stdout ----
>  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CargoError { cause: Some(NotFoundError { path: "/pyrsia/target/debug/pyrsia" }) }', pyrsia_cli/src/main.rs:85:52
>  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Root cause is `cargo tarpaulin` is not generating the following. Due to that CLI unit testing using command is failing.

```shell
`-- target
    |-- debug
    |   |-- libpyrsia.d
    |   |-- libpyrsia.rlib
    |   |-- libpyrsia_blockchain_network.d
    |   |-- libpyrsia_blockchain_network.rlib
    |   |-- pyrsia
    |   |-- pyrsia.d
    |   |-- pyrsia_node
    |   `-- pyrsia_node.d
```

.github/workflows/rust.yml
```diff
@@ -288,11 +288,15 @@ jobs:
...
+      run: |
+        cargo clean
+        cargo build --all-targets --workspace
+      shell: /bin/bash -e {0}
 
     - name: Run cargo-tarpaulin
       uses: actions-rs/tarpaulin@v0.1
       with:
-        args: '--all-targets --workspace -- --test-threads 1'
+        args: '--skip-clean --all-targets --workspace -- --test-threads 1'
```
dasmanas, line 299: `cargo tarpaulin` doesn't build binaries under `target/debug`. Creating binaries under `target/debug` using regular cargo build and using cargo tarpaulin with option `--skip-clean` to preserve the created binaries from the previous step so that CLI unit test can be performed.

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -53,13 +54,25 @@ pub fn config_add() {
...
+    match config::get_config_file_path() {
+        Ok(path_buf) => {
+            println!(
+                "{} {}",
+                CONF_FILE_PATH_MSG_STARTER,
+                path_buf.into_os_string().into_string().unwrap()
+            )
+        }
+        Err(error) => {
+            println!("Error retrieving config file path: {}", error);
```
karolh2000, line 66: Is it possible to be less aggressive with the error comment, are we able to distinguish if the file is not created yet (not sure if this is an option)?  For example if the file doesn't exist maybe something like "The config file not found".

karolh2000 (approved):

pyrsia_cli/src/cli/handlers.rs
dasmanas: Very valid question. We are using `confy::load` and it states `A new configuration file is created with default values if none exists.` And it is working as expected. Default is defined under `pyrsia/src/cli_commands/config.rs` file. 
```
impl Default for CliConfig {
    fn default() -> Self {
        CliConfig {
            host: "localhost".to_string(),
            port: "7888".to_string(),
            disk_allocated: "5.84 GB".to_string(),
        }
    }
}
```
I tested the same scenario. Deleted the file from the location and run `pyrsia config -show` and it created the file with the default value as defined.

.github/workflows/rust.yml
```diff
@@ -289,10 +289,18 @@ jobs:
...
+      run: |
+        cargo clean
+        cargo build --all-targets --workspace
+      shell: /bin/bash -e {0}
+
     - name: Run cargo-tarpaulin
       uses: actions-rs/tarpaulin@v0.1
       with:
-        args: '--all-targets --workspace -- --test-threads 1'
+        args: '--skip-clean --all-targets --workspace -- --test-threads 1'
```
fishseabowl, line 303: Can you provide more details about `skip-clean`? Thanks

fishseabowl (approved):

karolh2000 (approved):

PR #1261, opened 2022-11-02
Enable debug on kubernetes pod

sbtaylor15:
## Description

Enable debug as the default for the k8s environment.

Fixes pyrsia/pyrsia#

This PR enables RUST_LOG=DEBUG by setting the environment variable at the deployment level which is then passed on to the container.

## PR Checklist

- [X] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/community/get_involved/good_pr.md)

### Code Contributions

### DIFF
--- a/installers/helm/pyrsia-node/templates/deployment.yaml
+++ b/installers/helm/pyrsia-node/templates/deployment.yaml
@@ -31,6 +31,8 @@ spec:
             - name: p2p
               containerPort: 44000
           env:
+            - name: RUST_LOG
+              value: DEBUG
             - name: PYRSIA_ARTIFACT_PATH
               value: /usr/local/var/pyrsia
             - name: PYRSIA_BLOCKCHAIN_PATH 


efrisch (approved):
Does what it says: Turns on debug. 👍

installers/helm/pyrsia-node/templates/deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
             - name: p2p
               containerPort: 44000
           env:
+            - name: RUST_LOG
+              value: DEBUG
```
dasmanas, line 35: Should we consider DEBUG logging across the board
or 
`RUST_LOG="info,pyrsia=debug"`

dasmanas, line 35:
```suggestion
            - name: RUST_LOG
              value: info,pyrsia=debug
```

dasmanas: in that case `debug` will be set only to pyrsia modules instead of across all the modules.

sbtaylor15: Got it.  Going to leave debug turned on for all modules so we gather as much debug data we can.  We will need to turn it back to INFO when we are more stable.

Issue #1273, opened 2022-11-09
Running tests modify CLI config settings


jperedadnr:
### Steps taken/Steps to reproduce

Remove  ` /Users/$user/Library/Preferences/rs.pyrsia-cli/`, clean/build Pyrsia, and check the configuration, a default configuration is created:

```
$ ./pyrsia config -s
Config file path: /Users/$user/Library/Preferences/rs.pyrsia-cli/default-config.toml
host = 'localhost'
port = '7888'
disk_allocated = '5.84 GB'
```

Modify as needed:
```
$ ./pyrsia config -e
Enter host: 
localhost
Enter port: 
7889
Enter disk space to be allocated to pyrsia(Please enter with units ex: 10 GB): 
10 GB
Node configuration Saved !!
```

Verify it:
```
./pyrsia config -s
Config file path: /Users/$user/Library/Preferences/rs.pyrsia-cli/default-config.toml
host = 'localhost'
port = '7889'
disk_allocated = '10 GB'
```

Now run tests:
```
cargo test --workspace
```

And check the configuration again, it has been changed to the values that are set in `cli/handlers.rs::test_valid_config_edit` test:
```
$ ./pyrsia config -s
Config file path: /Users/$user/Library/Preferences/rs.pyrsia-cli/default-config.toml
host = 'some.localhost'
port = '65535'
disk_allocated = '10 GB'
```

### Expected Behavior

configuration remains the same as it was before running tests.

### Experienced Instance of the bug

rustc 1.63.0 (4b91a6ea7 2022-08-08)
Mac

betarelease:
Related and fixed in #1279

PR #1279, opened 2022-11-10
Keep the config intact as previous state

dasmanas:
## Description

Fixes pyrsia/pyrsia#1273

This PR keeps the config in original state after test case run completed.

## PR Checklist

### Code Contributions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1567,9 +1567,9 @@ dependencies = [
 
 [[package]]
 name = "httparse"
-version = "1.7.1"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "496ce29bb5a52785b44e0f7ca2847ae0bb839c9bd28f69acac9b99d461c0c04c"
+checksum = "d897f394bad6a705d5f4104762e116a75639e470d80901eed05a860a95cb1904"
 
 [[package]]
 name = "httpdate"
@@ -1616,9 +1616,9 @@ checksum = "9a3a5bfb195931eeb336b2a7b4d761daec841b97f947d34394601737a7bba5e4"
 
 [[package]]
 name = "hyper"
-version = "0.14.20"
+version = "0.14.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "02c929dc5c39e335a03c405292728118860721b10190d98c2a0f0efd5baafbac"
+checksum = "034711faac9d2166cb1baf1a2fb0b60b1f277f8492fd72176c17f3515e1abd3c"
 dependencies = [
  "bytes",
  "futures-channel",
@@ -2762,15 +2762,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19e64526ebdee182341572e50e9ad03965aa510cd94427a4549448f285e957a1"
 dependencies = [
  "hermit-abi",
- "libc",
-]
-
-[[package]]
-name = "num_threads"
-version = "0.1.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2819ce041d2ee131036f4fc9d6ae7ae125a3a40e97ba64d04fe799ad9dabbb44"
-dependencies = [
  "libc",
 ]
 
@@ -3970,9 +3961,9 @@ dependencies = [
 
 [[package]]
 name = "socket2"
-version = "0.4.4"
+version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66d72b759436ae32898a2af0a14218dbf55efde3feeb170eb623637db85ee1e0"
+checksum = "02e2d2db9033d13a1567121ddd7a095ee144db4e1ca1b1bda3419bc0da294ebd"
 dependencies = [
  "libc",
  "winapi",
@@ -4176,13 +4167,29 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.15"
+version = "0.3.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d634a985c4d4238ec39cacaed2e7ae552fbd3c476b552c1deac3021b7d7eaf0c"
+checksum = "a561bf4617eebd33bca6434b988f39ed798e527f51a1e797d0ee4f61c0a38376"
 dependencies = [
  "itoa",
- "libc",
- "num_threads",
+ "serde",
+ "time-core",
+ "time-macros",
+]
+
+[[package]]
+name = "time-core"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e153e1f1acaef8acc537e68b44906d2db6436e2b35ac2c6b42640fff91f00fd"
+
+[[package]]
+name = "time-macros"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d967f99f534ca7e495c575c62638eebc2898a8c84c119b89e250477bc4ba16b2"
+dependencies = [
+ "time-core",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4251,9 +4251,9 @@ dependencies = [
 
 [[package]]
 name = "tokio-stream"
-version = "0.1.9"
+version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df54d54117d6fdc4e4fea40fe1e4e566b3505700e148a6827e59b34b0d2600d9"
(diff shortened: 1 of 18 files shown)


Cargo.lock
```diff
@@ -4147,18 +4147,18 @@ checksum = "949517c0cf1bf4ee812e2e07e08ab448e3ae0d23472aee8a06c985f0c8815b16"
 
 [[package]]
 name = "thiserror"
-version = "1.0.35"
```
jperedadnr, line 4150: why the change here, and not in `Cargo.toml`?

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -357,6 +365,7 @@ mod tests {
 
     #[test]
     fn test_valid_config_edit() {
+        let existing_cli_config = config::get_config().unwrap();
```
jperedadnr, line 368: This works fine now, of course. 

But it might be a different option to use some different file and not the one "in production"?

So far we are using the same default file via `get_config()` that uses `confy::load(CONF_FILE, None)`, which in turn calls `config_name.into().unwrap_or("default-config")`, but that has room for improvement.

If we use for instance, `confy::load(CONF_FILE, TEST_FILE)`, then we won't override the user configuration.
Does this make sense to you?

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -365,13 +374,21 @@ mod tests {
...
         };
-        config_edit(host_name, port, diskspace);
+        let config_edit_result = config_edit(host_name, port, diskspace);
         let updated_cli_config = config::get_config().unwrap();
+        match config_edit_result {
+            Ok(_) => {
+                //restore the config to original state after test
+                let _restore_config = config::add_config(existing_cli_config);
+            }
+            Err(_) => {}
```
jperedadnr, line 384: If there is an error, shouldn't we also restore the original file? And shouldn't we log such error?

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -380,8 +397,15 @@ mod tests {
...
```
jperedadnr, line 407: Same here

Cargo.lock
dasmanas: I can see the file `src/blockchain/Cargo.toml` in the commit. It is at the end.

pyrsia_cli/src/cli/handlers.rs
dasmanas: In the recent latest version [`confy::load`](https://docs.rs/confy/latest/confy/fn.load.html) changed the function signature. It introduced a new arg `app_name`. The use case more like `confy::load("pyrsia", "cli")`, `confy::load("pyrsia", "node")` or `confy::load("pyrsia", "some_other_module")` instead of `confy::load(CONF_FILE, None)` what we are currently using.

pyrsia_cli/src/cli/handlers.rs
dasmanas: `config_edit_result` is holding the outcome of `pub fn config_edit(host_name: Option<String>, port: Option<String>, diskspace: Option<String>,) -> Result<(), Error>`. If there is error at the time of doing edit, we have nothing to restore back. Because anyway edit commit is not successful.

pyrsia_cli/src/cli/handlers.rs
dasmanas: `config_edit_result` is holding the outcome of `pub fn config_edit(host_name: Option<String>, port: Option<String>, diskspace: Option<String>,) -> Result<(), Error>`. If there is error at the time of doing edit, we have nothing to restore back. Because anyway edit commit is not successful.

Cargo.lock
jperedadnr: yes, but don't you need to change the main [`Cargo.toml` ](https://github.com/pyrsia/pyrsia/blob/main/Cargo.toml) at the root level? 
Imo, blockchain has nothing to do with this PR...

dasmanas: Reverted back the `Cargo.toml`. Change not required. `cargo clean` did the trick.

pyrsia_cli/src/cli/handlers.rs
dasmanas: new issue is created https://github.com/pyrsia/pyrsia/issues/1285

Cargo.lock
```diff
@@ -4156,18 +4156,18 @@ checksum = "949517c0cf1bf4ee812e2e07e08ab448e3ae0d23472aee8a06c985f0c8815b16"
...
+version = "1.0.37"
```
jperedadnr, line 4159: Don't you need to reverse these changes too?

pyrsia_cli/src/cli/handlers.rs
jperedadnr: Then you can skip the `match` and simply use
```
if config_edit_result.is_ok() {
   // restore the ...
}
```

pyrsia_cli/src/cli/handlers.rs
jperedadnr: same here

Cargo.lock
dasmanas: interesting thing is `Cargo.lock` is not getting change even though I change the `Cargo.toml`. I shouldn't change it manually too. Let me see what is going on.

dasmanas: `thiserror = "1.0.35"` means the range is `>=1.0.35, <2.0.0`. Explained under [Specifying dependencies from crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)

**Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows [SemVer](https://semver.org/) compatible updates. An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping.**

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -93,14 +98,17 @@ pub fn config_edit(host_name: Option<String>, port: Option<String>, diskspace: O
...
+                Ok(())
             } else {
                 errors.into_iter().for_each(|x| println!("{}", x));
-            }
+                Err(anyhow!("Invalid pyrsia config"))
+            };
         }
         Err(error) => {
             println!("Error Saving Node Configuration:       {}", error);
+            Err(anyhow!("Error Saving Node Configuration:       {}", error))
```
tiainen, line 109: Instead of both printing and returning the error, we can just return the error and print it where the `config_edit` method is invoked?

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -93,14 +98,17 @@ pub fn config_edit(host_name: Option<String>, port: Option<String>, diskspace: O
                         println!("Error Saving Node Configuration:       {}", error);
```
tiainen, line 98: Return this as an error too?

pyrsia_cli/src/cli/handlers.rs
dasmanas: changed as suggested.

pyrsia_cli/src/cli/handlers.rs
dasmanas: changed as suggested.

tiainen:
@dasmanas can you replace the `match` with an `if` as suggested by José? Other than that, the PR looks fine to me.

dasmanas:
> @dasmanas can you replace the `match` with an `if` in the tests as [suggested by José](https://github.com/pyrsia/pyrsia/pull/1279/files#r1020539779)? Other than that, the PR looks fine to me.

changed as per suggestion.

pyrsia_cli/src/cli/handlers.rs
dasmanas: `if config_edit_result.is_ok()` done

pyrsia_cli/src/cli/handlers.rs
dasmanas: if config_edit_result.is_ok()

dasmanas:
@tiainen , @jperedadnr PR is ready for review.

codecov/patch — 36.36% of diff hit (target 47.50%) -> This only can be fixed with the introduction of `mockall` in pyrsia code base and some code refactoring is required. As some bare functions need to go under `struct` or `impl` to mock behavior or outcome

Cargo.lock
jperedadnr: You did revert the `src/blockchain/Cargo.toml` file, that was fine.

But why `Cargo.lock` is still modified? That would happen if you modify the main `Cargo.toml` with:

```
-thiserror = "1.0.35"
+thiserror = "1.0.37"
``` 

Can you please either apply this change to `Cargo.toml` or reset the current change from `Cargo.lock`?

dasmanas: Done.

tiainen (approved):

jperedadnr (approved):

Issue #1246, opened 2022-10-28
Artifact p2p provide should happen before new block broadcast


erwin1:
Current situation:

A build node starts a build for a new artifact. When the build succeeds, it follows this procedure:

1. it creates the transparency log for the artifact and uses the transparency log service to add it to the ledger
2. the transparency log service uses the blockchain service to add the log as a payload to a new block
 (this is where the consensus should happen)
3. when the block is added, it is broadcast to connected peers, who store the transparency log in their local log database
4. after this, the log is written to the build node's transparency log database
5. and only after this, the artifact file is provided on the p2p network

The problem with this:

- step 3 is awaits the broadcast of new blocks, so step 4 and 5 only happens when the broadcast is done
- as a result, a peer that receives the block early on, can have a transparency log for a given artifact, but it won't be able to download it because step 5 didn't execute yet

Possible solution:

- the broadcast of new blocks should not block the publication process
- explore better ways for this mechanism to work, and we should include the consensus mechanism in our thinking because this might change the way this works anyway

efrisch:
One recommendation from the libp2p team was implementing the [ping protocol](https://github.com/libp2p/rust-libp2p/blob/master/protocols/ping/src/lib.rs). Which is more of a heartbeat in libp2p. From [here](https://docs.libp2p.io/concepts/fundamentals/protocols/),
"The ping protocol is a simple liveness check that peers can use to quickly see if another peer is online."

dasmanas:
Hey @erwin1 is this issue still relevant? I put together a flow which is currently implemented inside handle build result.

![HandleBuildResult](https://user-images.githubusercontent.com/12514131/215853662-e25950f6-d9aa-47a2-aac7-0164412b6d7b.png)

Link to the diagram: https://lucid.app/lucidchart/ac15c21f-1e25-42bf-b672-2fac99cd1e1b/edit?viewport_loc=-379%2C60%2C2335%2C1095%2C0_0&invitationId=inv_a2ce1aed-a58c-4e92-aeee-9e4d3546925a

PR #1284, opened 2022-11-10
Broadcast blockchain in a non-blocking way

jperedadnr:
## Description

Fixes pyrsia/pyrsia#1246

This PR changes `blockchain_service/service::broadcast_blockchain` implementation to make it asynchronous, so `add_artifact` or `add_authorized_node` in TransparencyLogService are not blocked anymore by this step.

The test `test_notify_blockchain` has been modified accordingly.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/blockchain_service/service.rs
+++ b/src/blockchain_service/service.rs
@@ -17,7 +17,7 @@
 use bincode::{deserialize, serialize};
 use libp2p::identity;
 use libp2p::PeerId;
-use log::warn;
+use log::{debug, warn};
 use pyrsia_blockchain_network::blockchain::Blockchain;
 use pyrsia_blockchain_network::error::BlockchainError;
 use pyrsia_blockchain_network::structures::block::Block;
@@ -116,7 +116,7 @@ impl BlockchainService {
         Ok(())
     }
 
-    /// Notify other nodes to add a new block.
+    /// Notify other nodes to add a new block, in a non-blocking way.
     async fn broadcast_blockchain(&mut self, block: Box<Block>) -> Result<(), BlockchainError> {
         let peer_list = self.p2p_client.list_peers().await.unwrap_or_default();
         let cmd = BlockchainCommand::Broadcast as u8;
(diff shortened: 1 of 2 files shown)


fishseabowl (dismissed):
lgtm

src/blockchain_service/service.rs
```diff
@@ -116,7 +116,7 @@ impl BlockchainService {
         Ok(())
     }
 
-    /// Notify other nodes to add a new block.
+    /// Notify other nodes to add a new block, in a non-blocking way.
     async fn broadcast_blockchain(&mut self, block: Box<Block>) -> Result<(), BlockchainError> {
```
tiainen, line 120: Should we consider using a void method instead of returning a Result? The alternative is to replace the `unwrap` methods that are used inside this method by returning an error (instead of panicking or using the default value).

src/blockchain_service/service.rs
```diff
@@ -456,15 +465,63 @@ mod tests {
...
         let block = Box::new(Block::new(
             HashDigest::new(b""),
             1,
             vec![],
             &blockchain_service.keypair,
         ));
-        assert!(blockchain_service.broadcast_blockchain(block).await.is_ok());
+
+        let (_, result) = tokio::join!(
+            tokio::spawn(async move {
```
tiainen, line 489: There's no real need for spawning separate threads, you can pass along a future as well.

```
let (_, result) = tokio::join!(
    blockchain_service.broadcast_blockchain(block),
    async {
        let value: u8;
        loop {
            ...
        }
        value
    }
)
```

src/blockchain_service/service.rs
```diff
@@ -456,15 +465,63 @@ mod tests {
...
+                loop {
+                    match receiver.recv().await {
+                        Some(Command::ListPeers { sender, .. }) => {
+                            let mut set = HashSet::new();
+                            set.insert(local_peer_id);
+                            if let Err(e) = sender.send(set) {
+                                panic!("Error sending peer id: {:?}", e);
+                            }
+                        }
+                        Some(Command::RequestBlockchain {
```
tiainen, line 506: If you don't use some of the struct parameters in a match, you can leave them out and replace them with `..`, just like the first match above:

```
    Some(Command::RequestBlockchain { data, .. }) => {
        ...
    }
```

src/blockchain_service/service.rs
jperedadnr: Not sure if I understand, what would be the benefit of not returning a Result in this case?

src/blockchain_service/service.rs
jperedadnr: thanks, fixed

src/blockchain_service/service.rs
jperedadnr: thanks, fixed

erwin1:
I'm having intermittent errors when running multiple nodes in a test:

> This node failed to update blockchain Error: Lagging Blockchain Data

it probably happens because we trigger request_blockchain in separate threads now, which may influence order.

erwin1:
it was decided to decline this solution for now: https://github.com/pyrsia/pyrsia/issues/1246#issuecomment-1322277895

PR #1309, opened 2022-11-18
Proper Disk Space Validation

dasmanas:
## Description

Fixes pyrsia/pyrsia#

This PR does... by accomplishing... and it can be reviewed by... you can also test the changes by running...

## PR Checklist

- [ ] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/community/get_involved/good_pr.md)

### Code Contributions

- [ ] I've made sure my rust toolchain is current `rustup update`.

### DIFF
--- a/pyrsia_cli/src/cli/handlers.rs
+++ b/pyrsia_cli/src/cli/handlers.rs
@@ -288,19 +288,19 @@ fn valid_port(input: &str) -> bool {
 }
 
 fn valid_disk_space(input: &str) -> bool {
-    const DISK_SPACE_NUM_MIN: u16 = 0;
-    const DISK_SPACE_NUM_MAX: u16 = 4096;
+    const DISK_SPACE_NUM_MIN: f32 = 0 as f32;
+    const DISK_SPACE_NUM_MAX: f32 = 4096 as f32;
     lazy_static! {
-        static ref DISK_SPACE_RE: Regex = Regex::new(r"^([0-9]{1,4})\s+(GB)$").unwrap();
+        static ref DISK_SPACE_RE: Regex =
+            Regex::new(r"^(([0-9]{1,4})((\.){1}([0-9]{1,3})){0,1})\s+(GB)$").unwrap();
     }
-    // let disk_space_re: Regex = Regex::new(r"^([0-9]{1,4})(\s*)(GB)$").unwrap();
     if DISK_SPACE_RE.is_match(input) {
         let captured_groups = DISK_SPACE_RE.captures(input).unwrap();
         let disk_space_num = captured_groups
             .get(1)
             .unwrap()
             .as_str()
-            .parse::<u16>()
+            .parse::<f32>()
             .unwrap();
         DISK_SPACE_NUM_MIN < disk_space_num && disk_space_num <= DISK_SPACE_NUM_MAX
     } else {
@@ -347,13 +347,23 @@ mod tests {
 
     #[test]
     fn test_valid_disk_space() {
-        let valid_disk_space_list = vec!["100 GB", "1 GB", "4096 GB"];
+        let valid_disk_space_list = vec![
+            "100 GB", "1 GB", "4096 GB", "0.22 GB", "1.22 GB", "0.22 GB", "5.84 GB",
+        ];
         assert!(valid_disk_space_list.into_iter().all(valid_disk_space));
     }
 
     #[test]
     fn test_invalid_disk_space() {
-        let invalid_disk_space_list = vec!["0 GB", "4097 GB", "100GB", "100gb"];
+        let invalid_disk_space_list = vec![
+            "0 GB",
+            "4097 GB",
+            "100GB",
+            "100gb",
+            "5..84 GB",
+            "5..84 GB",
+            "5.84.22 GB",
+        ];
         assert!(!invalid_disk_space_list.into_iter().any(valid_disk_space));
     }
 


pyrsia_cli/src/cli/handlers.rs
```diff
@@ -288,19 +288,19 @@ fn valid_port(input: &str) -> bool {
...
 
 fn valid_disk_space(input: &str) -> bool {
-    const DISK_SPACE_NUM_MIN: u16 = 0;
-    const DISK_SPACE_NUM_MAX: u16 = 4096;
+    const DISK_SPACE_NUM_MIN: f32 = 0 as f32;
+    const DISK_SPACE_NUM_MAX: f32 = 4096 as f32;
     lazy_static! {
-        static ref DISK_SPACE_RE: Regex = Regex::new(r"^([0-9]{1,4})\s+(GB)$").unwrap();
+        static ref DISK_SPACE_RE: Regex =
+            Regex::new(r"^(([0-9]{1,4})((\.){1}([0-9]{1,3})){0,1})\s+(GB)$").unwrap();
```
ihcomega56, line 295: This accepts values starting with 0 such as `001.2 GB` or `030 GB`.
It can be parsed to `f32`, but seems not proper.

ihcomega56 (changes requested):
Good update. Added one minor comment.

pyrsia_cli/src/cli/handlers.rs
betarelease: We should not need to parse decimals. We can stick with just whole numbers, and also only GB or higher allocation. Providing the ability to add decimal precision does not make sense given how much data we will need to store. Additionally it adds some much regex processing which is not easy to understand.

betarelease (commented):
I think this is not a bug and does not need fixing. As long as it supports whole numbers and GB or higher sizes we do not need to provide any more granularity. Adding this much regex makes code complex for not enough benefit.

pyrsia_cli/src/cli/handlers.rs
ihcomega56: Now that I think about it, @betarelease you are right. I guess users don't want to set decimals. Only we should do is tell users that whole numbers can be accepted via the CLI message.

Still this issue I pointed out remains, but shouldn't we care? If we don't merge this PR, values such as `0033 GB` can be set.

dasmanas: - code has been updated. Made the regex really simple just to capture in 2 groups.
  - Group 1: Capturing number and decimal character
  - Group 2: disk space metric unit i.e. GB
- if you provide 0001.99000 GB, it is going to save 1.99 GB in config file

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -274,37 +287,52 @@ fn valid_host(input: &str) -> bool {
...
 }
 
-fn valid_disk_space(input: &str) -> bool {
+/// Disk space will only accept integer values. Currently we it will accept value greater than 0 GB till 4096 GB
+fn valid_disk_space(input: &str) -> Result<String, Error> {
     const DISK_SPACE_NUM_MIN: u16 = 0;
     const DISK_SPACE_NUM_MAX: u16 = 4096;
     lazy_static! {
-        static ref DISK_SPACE_RE: Regex = Regex::new(r"^([0-9]{1,4})\s+(GB)$").unwrap();
+        static ref DISK_SPACE_RE: Regex = Regex::new(r"^([0-9,\.]+)\s+(GB)$").unwrap();
```
ihcomega56, line 314: It is more clear to reject decimals here by regex.

ihcomega56:
Thank you for updating. Added one comment.

pyrsia_cli/src/cli/handlers.rs
ihcomega56: @dasmanas Should we update it too?

dasmanas: I kept it intentionally because in future if we decide to accept float as well, we just need to change the parse statement  `let float_parsed_rslt = captured_groups.get(1).unwrap().as_str().parse::<u16>();`

docs/community/get_involved/local_dev_setup.md
```diff
@@ -140,7 +140,7 @@ You can use the Pyrsia CLI to ensure that the peers are connected.
     $ ./pyrsia -s
     Connected Peers Count:       1
     Artifacts Count:             3 {"manifests": 1, "blobs": 2}
-    Total Disk Space Allocated:  5.84 GB
+    Total Disk Space Allocated:  10 GB
     Disk Space Used:             0.0002%
```
fishseabowl, line 144: Is this document outdated? After running `./pyrsia -s`, I only get 'Connected Peers Count:       1'.

dasmanas: Even I found that this page is outdated. Still made a change only to the section of `Total Disk Space Allocated:`. Let me create a separate PR for the same.

dasmanas: Created an issue https://github.com/pyrsia/pyrsia/issues/1319

fishseabowl (dismissed):

tiainen (dismissed):

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -274,37 +287,52 @@ fn valid_host(input: &str) -> bool {
...
+                ))
+            } else {
+                Err(anyhow!("Invalid value for Disk Allocation"))
+            }
+        } else {
+            Err(anyhow!("Invalid value for Disk Allocation"))
+        }
     } else {
-        false
+        Err(anyhow!("Invalid value for Disk Allocation"))
```
ihcomega56, line 335: This function also can be written like follows to avoid repeating this `else { Err(anyhow!...` part.
(I was curious and got a lecture from @efrisch . Thank you @efrisch.)

```rust
fn valid_disk_space(input: &str) -> anyhow::Result<String> {
    const DISK_SPACE_NUM_MIN: u16 = 0;
    const DISK_SPACE_NUM_MAX: u16 = 4096;
    lazy_static! {
        static ref DISK_SPACE_RE: Regex = Regex::new(r"^([0-9,\.]+)\s+(GB)$").unwrap();
    }
    if DISK_SPACE_RE.is_match(input) {
        let captured_groups = DISK_SPACE_RE.captures(input).unwrap();
        //Group 1 is numeric part including decimal & Group 2 is metric part
        let disk_space_num = captured_groups.get(1).unwrap().as_str().parse::<u16>()?;
        if DISK_SPACE_NUM_MIN < disk_space_num && disk_space_num <= DISK_SPACE_NUM_MAX {
            let disk_space_metric = captured_groups.get(2).unwrap();
            return Ok(format!("{disk_space_num} {disk_space_metric:?}"));
        }
    }
    Err(anyhow!("Invalid value for Disk Allocation"))
}
```

dasmanas: very good suggestion. cyclomatic complexity reduced.

Didn't implement `let disk_space_num = captured_groups.get(1).unwrap().as_str().parse::<u16>()?;` because with `?` the message of the error will be from parse error if any instead of logic control. Rather keeping the if block with 
```
if float_parsed_rslt.is_ok() {
// code
}
```

erwin1 (approved):

tiainen (approved):

ihcomega56 (approved):

dasmanas:
Separate issue got created to solve the linting issue https://github.com/pyrsia/pyrsia/issues/1326

PR #1315, opened 2022-11-22
Updates pull request template

betarelease:
This PR updates the template to not require each PR to run rustup update. This has broken stuff in the past and is agressive in terms of updating rust. Instead we would like to update rust more deliberately and diligently.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -42,4 +42,3 @@ For a complete list of steps, check out the [developer PR guidlines](https://git
 
 - [ ] I've built the code `cargo build --all-targets` successfully.
 - [ ] I've run the unit tests `cargo test --workspace` and everything passes.
-- [ ] I've made sure my rust toolchain is current `rustup update`.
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -42,4 +42,3 @@ For a complete list of steps, check out the [developer PR guidlines](https://git
 
 - [ ] I've built the code `cargo build --all-targets` successfully.
```
dasmanas, line 43: should we also include `--workspace`?

.github/pull_request_template.md
```diff
@@ -42,4 +42,3 @@ For a complete list of steps, check out the [developer PR guidlines](https://git
...
 - [ ] I've run the unit tests `cargo test --workspace` and everything passes.
```
dasmanas, line 44: should we also include `--no-fail-fast`?

betarelease: maybe in a separate PR if we really wanted to do this. The rustup change I made is specific because we dont want people to run rustup before running PRs.(or on their own anyway)

.github/pull_request_template.md
betarelease: maybe in a separate PR if we really wanted to do this. The rustup change I made is specific because we dont want people to run rustup before running PRs.(or on their own anyway)

dasmanas (approved):

fishseabowl (approved):
lgtm

Issue #1352, opened 2022-11-30
Use a fixed rust toolchain


erwin1:
This is related to #1288 but not the same.

As it turns out, we are NOT using a fixed toolchain version, so we have been using the version that github provided by default.
This is also the reason clippy started to fail in #1326 (fixed in #1341)

So the first step, before upgrading to 1.65.0 in #1288 is to allow our workflows to use a fixed toolchain version.

betarelease:
Toolchain specification
Standard release channel toolchain names have the following form:


<channel>[-<date>][-<host>]

<channel>       = stable|beta|nightly|<major.minor>|<major.minor.patch>
<date>          = YYYY-MM-DD
<host>          = <target-triple>





- name: Install stable toolchain
      uses: actions-rs/toolchain@v1
      with:
        toolchain: stable
        override: true

betarelease:
Also in the cargo.toml we can now specify rust-version for the package

https://doc.rust-lang.org/cargo/reference/manifest.html due to this RFC - https://github.com/rust-lang/rfcs/pull/1953 So that would be the way to go and specify `major.minor`

betarelease:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

erwin1:
it will be key to find all places where we use rust tools

e.g. there's this line in `installers/docker/CodeCoverage.Dockerfile`
```
FROM rust:1.57.0-buster
```

PR #1358, opened 2022-12-01
pin rust to 1.65 and display rust versions

sbtaylor15:
## Description

Explicitly set the tool chain version to 1.65.  Added a steps to display the rust versions to the log output.  Moved `strip` from a separate step to be part of the Release Profile in the config.toml so all OS perform strip as part of a Release Build in the GH Actions.

Fixes pyrsia/pyrsia#1352

## PR Checklist

- [X] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/community/get_involved/good_pr.md)

### Code Contributions

### DIFF
--- a/.github/workflows/rust-lint.yml
+++ b/.github/workflows/rust-lint.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: sbtaylor15/rust-toolchain@v1      
+      - uses: pyrsia/rust-toolchain@v1      
       - run: mkdir -p $HOME/.cargo
       - run: cp .github/workflows/.cargo/audit.toml $HOME/.cargo
       - uses: actions-rs/audit-check@v1
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: sbtaylor15/rust-toolchain@v1 
+      - uses: pyrsia/rust-toolchain@v1 
         with:
           components: rustfmt  
       - run: cargo fmt --version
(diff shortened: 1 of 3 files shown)


.github/workflows/rust-lint.yml
```diff
@@ -13,8 +13,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - name: Install 1.65 toolchain
+        id: toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.65
```
erwin1, line 20: could we use some kind of mechanism to re-use this step so we can easily change the version at one place?

erwin1:
We should also take care of all the base docker images we use that include rust:
e.g.:
https://github.com/pyrsia/pyrsia/blob/main/installers/docker/local_code_coverage/Dockerfile#L1
https://github.com/pyrsia/pyrsia/blob/main/Dockerfile#L3

I think we need a section in the developer docs to list all the steps to take when upgrading the rust version

dasmanas (commented):
Can we use environment variable in .yml file and keep on using the same? It will make the change easier in future. Reference [about-environment-variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables)

.github/workflows/rust-lint.yml
sbtaylor15: Added reusable [toolchain action](https://github.com/pyrsia/rust-toolchain)

Cargo.toml
```diff
@@ -7,6 +7,7 @@ description = "Pyrsia Node and CLI"
 authors = ["pyrsiaoss <pyrsiaopensource@gmail.com>"]
 license = "Apache-2"
 repository = "https://github.com/pyrsia/pyrsia"
+rust = "1.65"
```
karolh2000, line 10: Regarding our recent problems with Clippy, we should downgrade rust to 1.64 until 1.67 is stable (currently nightly)
```suggestion
rust = "1.64"
```

sbtaylor15: change to 1.64

Issue #1221, opened 2022-10-19
Show useful error messages in the Pyrsia CLI


erwin1:
When the Pyrsia CLI receives an error (e.g. http 400 or 500) it should display a better error message when possible.

Example: see #1215 where the build request fails because the artifact is already available ==> the pyrsia CLI should display a meaningful error, so the user does not have to view the logs just to see what is happening.

betarelease:
Once we trigger the build and if the build pipeline returns an error during the build process we just return the 400 error but do not have a callback to retrieve the details of the error.

erwin1:
moving to milestone 0.2.2

PR #1370, opened 2022-12-05
Return useful error messages in the Pyrsia CLI

ihcomega56:
## Description

Part of pyrsia/pyrsia#1221, but this PR does not fix everything.
After this PR is merged, the same change should be applied to all errors.

Details of this change are written in comments below.

## The current response
```
$ pyrsia build docker --image "alpine:3.16"                                                                                                                                          
Build request failed with error: HTTP status client error (400 Bad Request) for url (http://localhost:7888/build/docker)
```

## After this change
```
$ pyrsia build docker --image "alpine:3.16"                                                                                                                           
Build request failed with error: HTTP status error (400 Bad Request) for url (http://localhost:7888/build/docker): "Failed to start a build: Artifact ID alpine:3.16 for type Docker already exists in transparency log"
```

## PR Checklist

### Code Contributions

### DIFF
--- a/src/artifact_service/service.rs
+++ b/src/artifact_service/service.rs
@@ -104,7 +104,7 @@ impl ArtifactService {
         // prevent duplicated builds
         self.transparency_log_service
             .verify_package_can_be_added_to_transparency_logs(&package_type, &package_specific_id)
-            .map_err(|t| BuildError::ArtifactAlreadyExists(format!("{:?}", t)))?;
+            .map_err(|t| BuildError::ArtifactAlreadyExists(t.to_string()))?;
 
         if local_peer_id.eq(&peer_id) {
             debug!("Start local build in authorized node");
--- a/src/cli_commands/node.rs
+++ b/src/cli_commands/node.rs
@@ -14,11 +14,16 @@
    limitations under the License.
 */
 
-use super::config::get_config;
+use anyhow::anyhow;
+use reqwest::Response;
+use async_trait::async_trait;
+
 use crate::node_api::model::cli::{
     RequestAddAuthorizedNode, RequestDockerBuild, RequestDockerLog, RequestMavenBuild,
     RequestMavenLog, Status,
 };
+
+use super::config::get_config;
 
 pub async fn ping() -> Result<String, reqwest::Error> {
     //TODO: implement ping api in Node
@@ -52,19 +57,15 @@ pub async fn add_authorized_node(request: RequestAddAuthorizedNode) -> Result<()
         .map(|_| ())
 }
 
-pub async fn request_docker_build(request: RequestDockerBuild) -> Result<String, reqwest::Error> {
+pub async fn request_docker_build(request: RequestDockerBuild) -> Result<String, anyhow::Error> {
     let node_url = format!("http://{}/build/docker", get_url());
     let client = reqwest::Client::new();
-    match client
-        .post(node_url)
+    client.post(&node_url)
         .json(&request)
         .send()
         .await?
-        .error_for_status()
-    {
-        Ok(response) => response.json::<String>().await,
-        Err(e) => Err(e),
-    }
+        .error_for_status_with_body()
+        .await
 }
 
 pub async fn request_maven_build(request: RequestMavenBuild) -> Result<String, reqwest::Error> {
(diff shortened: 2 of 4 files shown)


src/docker/error_util.rs
```diff
@@ -157,6 +157,7 @@ pub async fn custom_recover(err: Rejection) -> Result<impl Reply, Infallible> {
             RegistryErrorCode::BadRequest(m) => {
                 status_code = StatusCode::BAD_REQUEST;
                 error_message.code = RegistryErrorCode::BadRequest(m.clone());
+                error_message.message = m.clone();
```
ihcomega56, line 160: The server currently returns JSON like this:
```json
[
  {
    "code": {
      "BadRequest": "Failed to start a build: Artifact ID alpine:3.16 for type Docker already exists in transparency log"
    },
    "message": ""
  }
]
```

This change is added to extract an error message by the following JSON
(If possible, it is better to remove `"BadRequest": "Failed to start a build: Artifact ID alpine:3.16 for type Docker already exists in transparency log"`, but it is going to be a big change, and may break backward compatibility).

```json
[
  {
    "code": {
      "BadRequest": "Failed to start a build: Artifact ID alpine:3.16 for type Docker already exists in transparency log"
    },
    "message": "Failed to start a build: Artifact ID alpine:3.16 for type Docker already exists in transparency log"
  }
]
```

src/cli_commands/node.rs
```diff
@@ -132,3 +132,19 @@ pub fn get_url() -> String {
 
     format!("{}:{}", host, port)
 }
+
+#[async_trait]
+trait ErrorResponseWithBody {
+    async fn error_for_status_with_body(self) -> Result<String, anyhow::Error>;
```
ihcomega56, line 138: It seems that reqwest  does not support an error type with response body (error messages)? 
That is why this function is added.
The way to parse the response may be improved though.

Also is the place to define this trait proper?

efrisch (approved):
lgtm. 👍

src/cli_commands/node.rs
```diff
@@ -52,34 +58,28 @@ pub async fn add_authorized_node(request: RequestAddAuthorizedNode) -> Result<()
         .map(|_| ())
 }
 
-pub async fn request_docker_build(request: RequestDockerBuild) -> Result<String, reqwest::Error> {
+pub async fn request_docker_build(request: RequestDockerBuild) -> Result<String, anyhow::Error> {
```
ihcomega56, line 61: More unit tests should be added in order to test this function. Now there is no test covering it.
I'm wondering how to do because this function sends a request to Pyrsia node (Maybe should I make use of tokio?). I'm happy if given some references such as the past PRs and other test cases which are similar.

src/artifact_service/service.rs
```diff
@@ -104,7 +104,7 @@ impl ArtifactService {
         // prevent duplicated builds
         self.transparency_log_service
             .verify_package_can_be_added_to_transparency_logs(&package_type, &package_specific_id)
-            .map_err(|t| BuildError::ArtifactAlreadyExists(format!("{:?}", t)))?;
+            .map_err(|t| BuildError::ArtifactAlreadyExists(t.to_string()))?;
```
fishseabowl, line 107: Could you explain the reason for changing the format macro to the to_string() function? Thanks

fishseabowl (approved):
lgtm

src/artifact_service/service.rs
ihcomega56: @fishseabowl
If I understand correctly, we should change it to return an error message written in `#[error()]` 
 ([code](https://github.com/pyrsia/pyrsia/blob/a1e6c6589d91c69314388b36b66e332811931580/src/transparency_log/log.rs#L47)).

A `Display` impl, which is used by `to_string()`, is created automatically by thiserror if we provide `#[error()]`.

### The difference of the response

### When we use format!
`Build request failed with error: HTTP status error (400 Bad Request) for url (http://localhost:7888/build/docker): "Failed to start a build: ArtifactAlreadyExists { package_type: Docker, package_specific_id: \"alpine:3.16.2\" }"
  `
It seems just the error type is formatted with its structure.

### When we use to_string()
`Build request failed with error: HTTP status error (400 Bad Request) for url (http://localhost:7888/build/docker): "Failed to start a build: Artifact ID alpine:3.16.2 for type Docker already exists in transparency log"`

It uses `#[error()]` effectivly.

ihcomega56: Thank you for the good question. Actually I didn't fully understand the reason 😓

ihcomega56:
Thank you for the review!

Issue #1320, opened 2022-11-22
build-osx workflow fails intermittently


dasmanas:
### Steps taken/Steps to reproduce

build-osx workflow fails intermittently. Cause is CLI Config. E.g. of failed workflow https://github.com/pyrsia/pyrsia/actions/runs/3527347821/jobs/5916302005.

Need to investigate following section of code from `src/cli_commands/config.rs`

```rust
impl Display for CliConfig {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        let config_toml = toml::to_string_pretty(&self).expect("toml format error");
        write!(f, "{}", config_toml)
    }
}
```

### Experienced Instance of the bug

rustc 1.63.0

dasmanas:
Able to replicate the issue in local machine. Some of our current test case can't run in parallel thread. Because the entire transaction during the test is not atomic.

For short term solution we have to pass `-- --test-threads=1` wherever we are using `cargo test` in our build pipeline in GitHub action.

PR #1376, opened 2022-12-07
Pyrsia CLI Config - individual attribute configuration issue

dasmanas:
## Description

Fixes pyrsia/pyrsia#1375, pyrsia/pyrsia#1320

This PR does following.
- Fixes issue introduced during `clap` crate upgrade in while updating config using single attribute.
  - A few more unit test cases added to safe guard CLI Config edit with single attribute.
- Some tests are failing in multithread scenario. This got fixed and will not fail in GH action anymore after the merge.
- Some functions consist of some core logic have moved to `src/cli_commands/config.rs`

## PR Checklist

- [x] I've read the [contributing guidelines](https://github.com/pyrsia/.github/blob/main/contributing.md).
- [x] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/community/get_involved/good_pr.md)
- [x] I've included a good title and brief description along with how to review them.
- [x] I've linked any associated an [issue](https://github.com/pyrsia/pyrsia/issues).

### Code Contributions

- [x] I've built the code `cargo build --all-targets` successfully.
- [x] I've run the unit tests `cargo test --workspace` and everything passes.

PR #1379, opened 2022-12-07
Run test in single thread

dasmanas:
## Description

Fixes pyrsia/pyrsia#1320

This PR will change GH Jobs OSX build & Linux build to run `cargo test` in single thread like it is currently doing in coverage job.

## PR Checklist

### Code Contributions

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,7 +28,7 @@ jobs:
       CARGO_INCREMENTAL: 0
     steps:
     - uses: actions/checkout@v3
-    - uses: pyrsia/rust-toolchain@v1 
+    - uses: pyrsia/rust-toolchain@v1
 
     # Use sscache store in GitHub cache
     - uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
     - name: Execute Test Cases
       run: |
-        cargo test --workspace --verbose --release
+        cargo test --workspace --verbose --release -- --test-threads=1
```
fishseabowl, line 64: Is it possible to not use a shared environment in unit tests? Thanks

dasmanas: > Is it possible to not use a shared environment in unit tests? Thanks

Can you please explain a little bit on the topic. One job executes in one runner and actions in a job runs sequentially in the same runner.  After the build is completed, test is running on the same runner. with the additional parameter `-- --test-threads=1` I am trying to make sure multiple test doesn't run concurrently.

fishseabowl: thank you for your reply. I agree with you that we can run all unit tests in one thread. But I think it is best that all unit tests are independent and do not affect each other. So we can run unit tests in parallel. thanks

dasmanas: Instead of making this change, I am working on the test cases so that they can be run at concurrently.

dasmanas:
will be closed by another PR https://github.com/pyrsia/pyrsia/pull/1376/

dasmanas:
https://github.com/pyrsia/pyrsia/pull/1376/

PR #1443, opened 2022-12-21
Enable tests on windows

### LINKED ISSUE #1094: test util::keypair_util::tests::keypair_generates_new_keypair_but_does_not_save_it fails on Windows

Run

```
cargo test --workspace
```

The test fails with:

```
---- util::keypair_util::tests::keypair_generates_new_keypair_but_does_not_save_it stdout ----
thread 'util::keypair_util::tests::keypair_generates_new_keypair_but_does_not_save_it' panicked at 'assertion failed: !tmp_file.as_path().exists()', src\util\keypair_util.rs:145:9
```

Adding debug info to the test:
```
println!("tmp file {:?} {}", tmp_file.as_path(), tmp_file.as_path().exists());
assert!(tmp_file.as_path().exists());
```

 On Mac:
```
tmp file "/var/folders/zx/s6qr37tn5l98xzzsxttp9lph0000gn/T/.tmpnNMunv/keypair_fails" false
```

but on Windows:

```
tmp file "C:\\Users\\$user\\AppData\\Local\\Temp\\.tmpF037IB\\keypair_fails" true
```

however I don't see it nor can't access it.

### Expected Behavior

The test doesn't fail

rust rustc 1.64.0 (a55dd71d5 2022-09-19)  
Windows 10

tiainen:
## Description

Fixes pyrsia#1094

This PR lets all tests run on the windows platform and updates the github workflow to enable running the tests on the windows runner.

In more detail, it covers the following changes:

* `cli_commands/config.rs`: don't use the `HOME` environment variable as it does not exist in windows
* `peer_metrics/metrics.rs`: use a single instance of `sysinfo::System` as [described in their documentation](https://docs.rs/sysinfo/0.27.1/sysinfo/#usage)
* `util/keypair_util.rs`: in windows you can still write into a read-only folder, use a non-existent path instead to test the failure case
* make sure the test resources are checked out with a LF line ending on all platforms

## PR Checklist

### Code Contributions


src/peer_metrics/metrics.rs
```diff
@@ -15,32 +15,61 @@
...
 const CPU_STRESS_WEIGHT: f64 = 2_f64;
 const NETWORK_STRESS_WEIGHT: f64 = 0.001_f64;
 const DISK_STRESS_WEIGHT: f64 = 0.001_f64;
 
-/// Get the local stress metric to advertise to peers
-pub fn get_quality_metric() -> f64 {
-    let mut qm = get_cpu_stress() * CPU_STRESS_WEIGHT;
-    qm += get_network_stress() * NETWORK_STRESS_WEIGHT;
-    qm + get_disk_stress() * DISK_STRESS_WEIGHT
+lazy_static::lazy_static! {
```
betarelease, line 28: Looks like this became a global static. Can we do without having this as global? Also does peer metrics need a mutex or could that be gathered without locking data structures?

src/cli_commands/config.rs
```diff
@@ -24,6 +24,27 @@ use regex::Regex;
 use serde::{Deserialize, Serialize};
 
 const CONF_FILE: &str = "pyrsia-cli";
+const ENV_VAR_PYRSIA_CONFIG_LOCATION_FOR_TEST: &str = "PYRSIA_CONFIG_LOCATION_FOR_TEST";
```
betarelease, line 27: This name constantization required some work for me to understand what this was for. But looks like the calls are to 

std::env::var(ENV_VAR..

so they might as well look like
std::env::("PYRSIA_CONFIG...") instead.

can we inject CONF_FILE where we need instead of having a separate conf path for testing?

src/peer_metrics/metrics.rs
tiainen: I've taken it out and created a single instance of PeerMetrics when starting the pyrsia node.

src/cli_commands/config.rs
tiainen: Renamed and added some documentation for clarification.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto
+
+tests/resources/*.json text eol=lf
```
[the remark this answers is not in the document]
tiainen: I can't seem to add the trialing newline. Do you have an idea what could be wrong?

fishseabowl (dismissed):
lgtm

betarelease (dismissed):

fishseabowl (approved):

erwin1 (approved):

Issue #1159, opened 2022-10-07
Automate Pyrsia node release procedure


erwin1:
not for 0.2 but for future releases

placeholder issue for release automation, let's try to record all steps we did manually for 0.2

erwin1:
- upload windows MSI to github release
- update documentation to include latest released version
- make sure apt repo and brew repo contain the correct latest release
- make sure the authorized nodes are updated

betarelease:
Subtask 
First document the test plan as described in #1339

erwin1:
* make sure to update the version in the Cargo.toml before tagging the release

erwin1:
The release procedure is now tracked in https://github.com/pyrsia/pyrsia/blob/main/docs/developers/release_process.md (more improvements in #1431)

Quick wins to improve this process are:
- Upload windows MSI to the GH release from within the action
- Use a variable for the version in rust.yml so it can be changed more easily
- Create a script to update and commit the new version number

PR #1444, opened 2022-12-21
update to next release version 0.2.3

erwin1:
Prepare for next release: set version to 0.2.3

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -130,7 +130,7 @@ jobs:
       - name: Add Build to Version Number
         shell: cmd
         run: |
-          "C:\Program Files\Git\usr\bin\sed.exe" -i "s/version = \"0.2.2\"/version = \"0.2.2+${{ github.run_number }}\"/" Cargo.toml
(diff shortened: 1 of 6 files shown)


.github/workflows/rust.yml
```diff
@@ -130,7 +130,7 @@ jobs:
       - name: Add Build to Version Number
         shell: cmd
         run: |
-          "C:\Program Files\Git\usr\bin\sed.exe" -i "s/version = \"0.2.2\"/version = \"0.2.2+${{ github.run_number }}\"/" Cargo.toml
+          "C:\Program Files\Git\usr\bin\sed.exe" -i "s/version = \"0.2.3\"/version = \"0.2.3+${{ github.run_number }}\"/" Cargo.toml
```
tiainen, line 133: Maybe we should change this at some point so it takes a regular expression, like it is already done for the macos runner. That way we don't need to update this each time we release a new version.

fishseabowl (approved):

.github/workflows/rust.yml
erwin1: yes indeed. this is captured in #1159

PR #1452, opened 2022-12-22
Initiate kademlia bootstrap after probe or dial

tiainen:
## Description

Fixes pyrsia#1446

This PR initiates the Kademlia DHT bootstrap after the first address of the boot node was added. In practice, this means that the bootstrap is started when either:

* a dial to another node has finished successfully
* another node was successfully added as the probe address for autonat

## PR Checklist

### Code Contributions

### DIFF
--- a/src/network/client.rs
+++ b/src/network/client.rs
@@ -78,7 +78,8 @@ pub struct Client {
 }
 
 impl Client {
-    // Add a probe address for AutoNAT discovery
+    /// Add a probe address for AutoNAT discovery. When adding the probe
+    /// was handled successfully, the kademlia DHT will be bootstrapped.
     pub async fn add_probe_address(
         &mut self,
         peer_id: &PeerId,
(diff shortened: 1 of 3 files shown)


karolh2000 (approved):

karolh2000:
looks good but it doesn't compile, I guess.

tiainen:
> looks good but it doesn't compile, I guess.

Indeed, I forgot to commit a file. It's all looking good now.

karolh2000 (approved):

src/network/client.rs
```diff
@@ -97,7 +98,15 @@ impl Client {
                 sender,
             })
             .await?;
-        receiver.await?
+        receiver.await??;
```
fishseabowl, line 101: I have a question: in this case, do we not handle any result other than error?

fishseabowl (approved):
lgtm

erwin1 (approved):

src/network/client.rs
tiainen: That's correct, both receivers receive the unit type () in case of success. Any error is propagated to the caller of this function.

PR #1464, opened 2022-12-30
Explicitly remove openssl before build if it is cached. Added dump of…

efrisch:
… cargo.lock if openssl is back.

## Description

Fixes pyrsia/pyrsia#1427

This PR adds an explicit removal of openssl if it is already cached in the cargo.lock and then it also dumps the cargo.lock if openssl is detected after the build.

## PR Checklist

- [x] I've read the [contributing guidelines](https://github.com/pyrsia/.github/blob/main/contributing.md).
- [x] I've read ["What is a Good PR?"](https://github.com/pyrsia/pyrsia/blob/main/docs/community/get_involved/good_pr.md)
- [x] I've included a good title and brief description along with how to review them.
- [x] I've linked any associated an [issue](https://github.com/pyrsia/pyrsia/issues).

PR #1467, opened 2022-12-30
1464b fail build if openssl returns

efrisch:
## Description

Fixes pyrsia/pyrsia#1464

This PR removes the hard exit on openssl, and removes the call to `cargo remove` (only available with nightly) and adds `cargo tree` when openssl is detected.

## PR Checklist

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -39,11 +39,11 @@ cat /tmp/tests.log
 echo "### Tests Build RC=$(cat /tmp/tests.rc)"
 
 # Check if OpenSSL is back
-if [[ $(find . -name "Cargo.lock" -exec grep -i openssl {} \; | wc -l) != 0 ]]; then
+if [[ $(find . -name "Cargo.lock" -exec grep -i openssl {} \; | wc -l) -ne 0 ]]; then
     echo "OpenSSL Presence detected in the Cargo; please remove it and rebuild. Dumping Cargo.lock files to log."
     find . -name "Cargo.lock" -exec cat {} \;
     cargo tree
-    # exit 1
+    exit 1
 fi
 
 # Return the max return code between the two processes.
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -40,8 +40,10 @@ echo "### Tests Build RC=$(cat /tmp/tests.rc)"
 
 # Check if OpenSSL is back
 if [[ $(find . -name "Cargo.lock" -exec grep -i openssl {} \; | wc -l) != 0 ]]; then
-    echo "OpenSSL Presence detected in the Cargo; please remove it and rebuild."
-    exit 1
+    echo "OpenSSL Presence detected in the Cargo; please remove it and rebuild. Dumping Cargo.lock files to log."
+    find . -name "Cargo.lock" -exec cat {} \;
+    cargo tree
+    # exit 1
 fi
 
 # Return the max return code between the two processes.
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -14,9 +14,6 @@
 
 set -e
 
-# Explicitly remove OpenSSL before build in case it is cached
-cargo remove openssl
-
 # We need to split out build commands into separate calls in order to run them in parallel
 # therefore we cannot use --all-targets since that is equivalent to --lib --bins --tests --benches --examples
 # which disables the parallelizm we are trying achive for speed. 


sbtaylor15 (dismissed):
/lgtm

betarelease:
Is the `build.sh` not callable to test this behaviour before we merge the PR?

.github/workflows/build.sh
```diff
@@ -45,7 +42,8 @@ echo "### Tests Build RC=$(cat /tmp/tests.rc)"
 if [[ $(find . -name "Cargo.lock" -exec grep -i openssl {} \; | wc -l) != 0 ]]; then
     echo "OpenSSL Presence detected in the Cargo; please remove it and rebuild. Dumping Cargo.lock files to log."
     find . -name "Cargo.lock" -exec cat {} \;
-    exit 1
+    cargo tree
```
fishseabowl, line 45: maybe `cargo tree | grep "openssl"` is better

efrisch:
> Is the `build.sh` not callable to test this behaviour before we merge the PR?

I get different behavior locally. For example, `git remove` works on my system. @sbtaylor15 feel free to chime in, is there a `dry-run` we can use?

.github/workflows/build.sh
efrisch: I thought about that. It would show openssl, but not what brought openssl in.

fishseabowl: Sure. Sometimes `cargo tree` doesn't show any openssl dependency.  What is the purpose of running`cargo tree`?

sbtaylor15:
@efrisch I think something like `cargo clean` will invalidate the caching completely causing longer builds.  `cargo update` may work to create an up to date Cargo.lock.

Or we leave the build process as it currently is and use `ldd` against the new executables to see if openssl is loaded as a runtime dependency.

.github/workflows/build.sh
sbtaylor15: This would stop the build and preserve caching.

```bash
# Display the tests return code
echo "### Tests Build RC=$(cat /tmp/tests.rc)"

# Return the max return code between the two processes.
# This is done to tell the GitHub Step to fail (rc != 0). The 
# value of the rc is not important at this point since each
# build has already displayed its return code. 

rc=$(cat /tmp/ws.rc /tmp/tests.rc | sort -nr | head -n1)

if ldd target/release/pyrsia_node | grep -q -i openssl; then
  echo "OpenSSL Presence detected in the pyrsia_node executable. Please remove it as a dependency and rebuild."
  exit 1
fi

if ldd target/release/pyrsia | grep -q -i openssl; then
  echo "OpenSSL Presence detected in the pyrsia executable. Please remove it as a dependency and rebuild."
  exit 1
fi

exit $rc
```

fishseabowl: Maybe checking `Cargo.lock` is better. I can't find the `openssl` using `otool -L`, but it appears in `Cargo.lock`. Please correct me if `ldd` is different from `otool -L`. Details are as follows:
```
xuefeihan@xuefeih-mac pyrsia % more Cargo.lock | grep "openssl"
 "openssl",
 "openssl-probe",
 "openssl-sys",
name = "openssl"
 "openssl-sys",
name = "openssl-probe"
name = "openssl-sys"
xuefeihan@xuefeih-mac pyrsia % otool -L ./target/debug/pyrsia_node
./target/debug/pyrsia_node:
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 60158.100.133)
	/System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1858.112.0)
	/System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 1163.100.19)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1858.112.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
```

sbtaylor15: @fishseabowl `ldd` is for Linux.

fishseabowl: > @fishseabowl `ldd` is for Linux.

Sure. `ldd` does not work on MacOS. `otool -L` is the equivalent command in MacOS.

sbtaylor15:
@efrisch `cargo build` will correctly update the Cargo.lock with the correct dependencies ie. add/remove them without changing the version numbers like `cargo update`.  The problem is in the if comparison.   

`if [[ $(find . -name "Cargo.lock" -exec grep -i openssl {} \; | wc -l) != 0 ]]; then` is a string comparison not a numeric comparison.  The wc -l will return '       0' in some cases and OSes.  The leading spaces cause the condition to be true  '    0' != '0', which makes it seem that openssl is present in the Cargo.lock.

The fix:

1. Remove `cargo remove openssl`
2. Change the if statement to:
```
# Check if OpenSSL is back
if [ $(find . -name "Cargo.lock" -exec grep -i openssl {} \; | wc -l) -ne 0 ]; then
    echo "OpenSSL Presence detected in the Cargo; please remove it and rebuild. Dumping Cargo.lock files to log."
    find . -name "Cargo.lock" -exec cat {} \;
    exit 1
fi
```

We should just fail the build for now and do another PR to do a notification instead.  I believe that Slack requires a webhook to be setup for us to post a message to our channels.

betarelease:
@sbtaylor15 if it comes as a normal notification from this repo we can use the subscribe feature and add it to our notifications channel. For webhooks it will require admin access and would become brittle because everytime we make a change this could break.

betarelease:
I would say let's stop this PR at failing the build when openssl is found and then let another PR take care of notification. Failing PR is the first step and the owner of the PR will need to take care of it.

sbtaylor15 (approved):
/lgtm

.github/workflows/build.sh
betarelease: I think this is the only conversation open on this. Once resolved and approved we should be able to merge. Is this conversation completed or do we need to make a fix?

fishseabowl (approved):

Issue #1399, opened 2022-12-13
Extended transparency log search


erwin1:
The goal is to have a fully flexible way of searching the transparency logs and the ability to output the result in different ways.

Let's start with what we currently have:

```
~ $ pyrsia inspect-log docker --help
pyrsia-inspect-log-docker
Show transparency logs for a Docker image

USAGE:
    pyrsia inspect-log docker --image <IMAGE>

OPTIONS:
        --image <IMAGE>    The docker image (e.g. alpine:3.15.3 or
                           alpine@sha256:1e014f84205d569a5cc3be4e108ca614055f7e21d11928946113ab3f36054801
    -h, --help             Print help information
```

So we can query the logs of one specific package type and for one specific artifact and output the all the logs we have in JSON format.

e.g. 

```
pyrsia inspect-log docker --image alpine:3.16.0
[
  {
    "artifact_hash": "2408cc74d12b6cd092bb8b516ba7d5e290f485d3eb9672efc00f0583730179e8",
    "artifact_id": "6df4ed3f-426a-4c74-8753-18e75ff3bbf2",
    "id": "220926df-7849-451e-a978-2ddd83a3d2fd",
    "node_id": "f2184738-b54f-4427-8fd9-446e7b4e17aa",
    "node_public_key": "be20fce8-d019-4ab4-8c2c-29776ab3804f",
    "num_artifacts": 4,
    "operation": "AddArtifact",
    "package_specific_artifact_id": "alpine@sha256:2408cc74d12b6cd092bb8b516ba7d5e290f485d3eb9672efc00f0583730179e8",
    "package_specific_id": "alpine:3.16.0",
    "package_type": "Docker",
    "source_hash": "",
    "source_id": "985b3dbd-cffc-436b-b361-a5c1b773ed40",
    "timestamp": 1666278549
  },
  {
    "artifact_hash": "4ff3ca91275773af45cb4b0834e12b7eb47d1c18f770a0b151381cd227f4c253",
    "artifact_id": "0c7b352d-9d4b-478a-9203-74b321bff1a9",
    "id": "e3ce63f8-e1cb-4a38-9719-c3ee690984f6",
    "node_id": "22919dd0-f7cd-4499-8a6f-a4443635abd8",
    "node_public_key": "f160a4ad-0c1d-4294-8c2b-490a4505873e",
    "num_artifacts": 4,
    "operation": "AddArtifact",
    "package_specific_artifact_id": "alpine:3.16.0",
[...]

PR #1525, opened 2023-01-18
CSV format to present `TransparencyLog` records

dgarus:
## Description

Part of #1399

- cli: added `--format <FORMAT> "The output format (json, csv")` to `inspect-log` command
- node_api: added `csv` format and tests for `/inspect/docker` and `inspect/maven` requests

## PR Checklist

### Code Contributions

### DIFF
--- a/pyrsia_cli/src/cli/handlers.rs
+++ b/pyrsia_cli/src/cli/handlers.rs
@@ -19,7 +19,7 @@ use pyrsia::cli_commands::config;
 use pyrsia::cli_commands::node;
 use pyrsia::node_api::model::cli::{
     RequestAddAuthorizedNode, RequestBuildStatus, RequestDockerBuild, RequestDockerLog,
-    RequestMavenBuild, RequestMavenLog,
+    RequestMavenBuild, RequestMavenLog, TransparentLogOutputParams,
 };
 use serde_json::Value;
 use std::collections::HashSet;
@@ -195,9 +195,10 @@ pub async fn node_list() {
     }
 }
 
-pub async fn inspect_docker_transparency_log(image: &str) {
+pub async fn inspect_docker_transparency_log(image: &str, format: Option<String>) {
     let result = node::inspect_docker_transparency_log(RequestDockerLog {
         image: image.to_owned(),
+        output_params: Some(TransparentLogOutputParams { format }),
     })
     .await;
     match result {
@@ -210,9 +211,10 @@ pub async fn inspect_docker_transparency_log(image: &str) {
     };
 }
 
-pub async fn inspect_maven_transparency_log(gav: &str) {
+pub async fn inspect_maven_transparency_log(gav: &str, format: Option<String>) {
     let result = node::inspect_maven_transparency_log(RequestMavenLog {
         gav: gav.to_owned(),
+        output_params: Some(TransparentLogOutputParams { format }),
     })
     .await;
     match result {
--- a/pyrsia_cli/src/cli/handlers.rs
+++ b/pyrsia_cli/src/cli/handlers.rs
@@ -28,6 +28,37 @@ use std::io::BufRead;
 
 const CONF_REMINDER_MESSAGE: &str = "Please make sure the pyrsia CLI config is up to date and matches the node configuration. For more information, run 'pyrsia config --show'";
 
+enum OutputFormat {
+    JSON,
+    CSV,
+}
+
+impl OutputFormat {
+    pub fn from(format: Option<&String>) -> Self {
+        if let Some(val) = format {
+            match val.as_str() {
+                "json" => OutputFormat::JSON,
+                "csv" => OutputFormat::CSV,
+                _ => panic!("Unknown format {:?}", val),
+            }
+        } else {
+            OutputFormat::JSON
+        }
+    }
+
+    pub fn print_logs(&self, logs: String) {
+        match self {
+            OutputFormat::JSON => {
+                let logs_as_json: Value = serde_json::from_str(logs.as_str()).unwrap();
+                println!("{}", serde_json::to_string_pretty(&logs_as_json).unwrap());
+            }
+            OutputFormat::CSV => {
+                println!("{}", logs);
+            }
+        }
+    }
+}
+
 pub fn config_add() -> anyhow::Result<()> {
     let default_config = config::CliConfig {
         ..Default::default()
@@ -196,14 +227,18 @@ pub async fn node_list() {
 }
 
 pub async fn inspect_docker_transparency_log(image: &str, format: Option<String>) {
+    let output_format = OutputFormat::from(format.as_ref());
+
     let result = node::inspect_docker_transparency_log(RequestDockerLog {
         image: image.to_owned(),
-        output_params: Some(TransparentLogOutputParams { format }),
+        output_params: Some(TransparentLogOutputParams {
+            format: format.clone(),
+        }),
     })
     .await;
     match result {
         Ok(logs) => {
-            print_logs(logs);
+            output_format.print_logs(logs);
         }
         Err(error) => {
             println!("Inspect log request failed with error: {:?}", error);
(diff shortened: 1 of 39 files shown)


fishseabowl (commented):
It looks good to me. I have only one concern: `unwrap()` will panic and crash the application. It would be better to replace it with another method of propagating errors if possible.

pyrsia_cli/src/cli/parser.rs
```diff
@@ -82,12 +82,14 @@ pub fn cli_parser() -> ArgMatches {
                         .arg_required_else_help(true)
                         .args(&[
                             arg!(--image <IMAGE> "The docker image (e.g. alpine:3.15.3 or alpine@sha256:1e014f84205d569a5cc3be4e108ca614055f7e21d11928946113ab3f36054801"),
+                            arg!(--format <FORMAT> "The output format (json, csv"),
```
ihcomega56, line 85: Why don't L.84 and 85 have a close braces `)`?

ihcomega56 (commented):
I run it on my machine and only 'json' seems to work. 'csv' returned an error, but let me know if I'm wrong. Also it is better to return useful error messages when a keyword other than 'json' or 'csv' is sent.

```
$ ./target/debug/pyrsia inspect-log docker --image alpine:3.16 --format csv
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 1)', pyrsia_cli/src/cli/handlers.rs:231:67
```

```
$ ./target/debug/pyrsia inspect-log docker --image alpine:3.16 --format test
Inspect log request failed with error: error sending request for url (http://localhost:7888/inspect/docker): connection closed before message completed

Caused by:
    connection closed before message completed
```

dgarus:
@ihcomega56 you are right, I should be more careful.
I fixed your comments, please, have a look.
Thanks!

src/node_api/handlers/swarm.rs
```diff
@@ -21,13 +21,79 @@ use crate::node_api::model::cli::{
...
+use crate::transparency_log::log::TransparencyLog;
 
 use crate::artifact_service::service::ArtifactService;
 use libp2p::PeerId;
 use log::debug;
 use std::str::FromStr;
 use warp::{http::StatusCode, Rejection, Reply};
 
+pub enum ContentType {
+    JSON(&'static str),
```
tiainen, line 33: I would remove the parameter on the enum types and add a method to extract the value for each type. That way, you don't always need to provide the value when constructing or using the type.

```
pub enum ContentType {
    JSON,
    CSV,
}

impl ContentType {
    pub fn value(&self) -> String {
        match self {
            ContentType::JSON => "application/json".to_owned(),
            ContentType::CSV => "text/csv".to_owned(),
        }
    }
}
```

Then, you can also add a default implementation for `ContentType`:
```
impl Default for ContentType {
    fn default() -> Self {
        ContentType::JSON
    }
}
```

src/node_api/model/cli.rs
```diff
@@ -33,9 +33,15 @@ pub struct RequestDockerBuild {
     pub image: String,
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct TransparentLogOutputParams {
```
tiainen, line 37: It's called transparency log.
```suggestion
pub struct TransparencyLogOutputParams {
```

src/node_api/model/cli.rs
```diff
@@ -33,9 +33,15 @@ pub struct RequestDockerBuild {
...
+    pub format: Option<String>,
```
tiainen, line 38: Define `format` as `ContentType` instead of `Option<String>`.
```suggestion
    pub content_type: ContentType,
```

src/node_api/model/cli.rs
```diff
@@ -33,9 +33,15 @@ pub struct RequestDockerBuild {
...
+}
+
 #[derive(Debug, Deserialize, Serialize)]
 pub struct RequestDockerLog {
     pub image: String,
+    pub output_params: Option<TransparentLogOutputParams>,
```
tiainen, line 44: I would not use `Option` here, but directly use the `TransparencyLogOutputParams`:
```suggestion
    pub output_params: TransparentLogOutputParams,
```

Keep the logic of handling the optional provisioning of the `format` argument in the CLI code.

src/node_api/model/cli.rs
```diff
@@ -46,9 +52,42 @@ pub struct RequestMavenBuild {
 #[derive(Debug, Deserialize, Serialize)]
 pub struct RequestMavenLog {
     pub gav: String,
+    pub output_params: Option<TransparentLogOutputParams>,
```
tiainen, line 55: Same here, no need for `Option`.

src/node_api/model/cli.rs
```diff
@@ -46,9 +52,42 @@ pub struct RequestMavenBuild {
...
 }
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct RequestBuildStatus {
     pub build_id: String,
 }
+
```
tiainen, line 62: All the changes below can be removed when directly using `ContentType` instead of `Option<String>` and when removing the Option from `Option<TransparencyLogOutputParams>`.

pyrsia_cli/src/cli/parser.rs
```diff
@@ -82,12 +82,14 @@ pub fn cli_parser() -> ArgMatches {
                         .arg_required_else_help(true)
                         .args(&[
                             arg!(--image <IMAGE> "The docker image (e.g. alpine:3.15.3 or alpine@sha256:1e014f84205d569a5cc3be4e108ca614055f7e21d11928946113ab3f36054801"),
+                            arg!(--format <FORMAT> "The output format (json, csv)"),
```
tiainen, line 85: I would make "json" the default value and explicitly define the possible values with the argument:
```suggestion
                            arg!(--format <FORMAT> "The output format")
                                .value_parser(["json", "csv"])
                                .default_value("json"),
```

pyrsia_cli/src/cli/parser.rs
```diff
@@ -82,12 +82,14 @@ pub fn cli_parser() -> ArgMatches {
...
                         ]),
                     Command::new("maven")
                         .about("Show transparency logs for a maven artifact")
                         .arg_required_else_help(true)
                         .args(&[
                             arg!(--gav <GAV> "The maven GAV (e.g. org.myorg:my-artifact:1.1.0)"),
+                            arg!(--format <FORMAT> "The output format (json, csv)"),
```
tiainen, line 92: Same here, use "json" as default value and specify possible values as above.

src/node_api/handlers/swarm.rs
```diff
@@ -21,13 +21,79 @@ use crate::node_api::model::cli::{
...
+
+impl ContentType {
+    pub fn from_string(format: Option<&String>) -> Self {
+        ContentType::from(match format {
+            Some(val) => Some(val.as_str()),
+            _ => None,
+        })
+    }
+
+    pub fn from(format: Option<&str>) -> Self {
```
tiainen, line 45: Implementing this method with the `FromStr` trait would be more idiomatic:

```
#[derive(Debug, PartialEq, Eq)]
pub struct ParseContentTypeError {
    invalid_type: String,
}

impl FromStr for ContentType {
    type Err = ParseContentTypeError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "json" => Ok(ContentType::JSON),
            "csv" => Ok(ContentType::CSV),
            _ => Err(ParseContentTypeError {
                invalid_type: s.to_owned(),
            }),
        }
    }
}
```

This allows you to use the `parse` method on any string:
```
    let format = "json";
    if let Ok(content_type) = format.parse::<ContentType>() {
        ...
    }
```

pyrsia_cli/src/cli/parser.rs
dgarus: great

src/node_api/handlers/swarm.rs
dgarus: cool

dgarus:
@tiainen thank you for the review!
I fixed your comments, please check.

We discussed with @tiainen using `Option` in `RequestDockerLog` (`TransparencyLogOutputParams`) and decided to keep them. It is useful for backward compatibility and allows omitting the `TransparencyLogOutputParams` part in requests.

tiainen (dismissed):

src/node_api/model/cli.rs
```diff
@@ -46,9 +53,46 @@ pub struct RequestMavenBuild {
...
+    }
+
+    fn format(params: &Option<Self>) -> swarm::ContentType {
+        if let Some(params) = params {
+            if let Some(val) = params.format {
+                return val;
+            }
+        };
+
+        swarm::ContentType::JSON
```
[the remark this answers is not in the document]
dgarus: Thanks! Sure.
I made 
```
Default::default()
```

tiainen (approved):

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -195,41 +195,46 @@ pub async fn node_list() {
     }
 }
 
-pub async fn inspect_docker_transparency_log(image: &str) {
+pub async fn inspect_docker_transparency_log(image: &str, format: Option<String>) {
+    let content_type = swarm::ContentType::from(format.as_ref()).unwrap();
```
fishseabowl, line 199: I have a simple question: what happens if `format` is `None`? Thanks

dgarus: We get default `ContentType` which is `JSON`.

fishseabowl (approved):
Good job. Thanks

Issue #817, opened 2022-06-29
Implement `get_authorized_nodes` in Transparency log service


erwin1:
The `get_authorized_nodes` method in the transparency log service needs to return a list of all current authorized nodes.

The transparency log service uses it database of transparency logs to do so.

First of all the TransparencyLog itself needs to expanded with extra fields to allow the recording of add_node and remove_node operations in the log:
- extra operations: add_node, remove_node
- extra field: node_id (the peer id of the authorized node)
- extra field: node_public_key (the public key of the authorized node)

(they will need to be included in the struct and in the database - how they are added is part of another issue)

This way we can define a struct AuthorizedNode with the following fields:
- id
- public_key

When `get_authorized_nodes` is called, the transparency log service will
- find all add_node and remove_node logs
- create a list of `AuthorizedNode`s that were added, and didn't get removed later
- return that list to the caller

PR #1532, opened 2023-01-20
Refactor `get_authorized_nodes()` method to return a list of PeerIDs

### LINKED ISSUE #1523: Implement the get_authorized_nodes method to return a list of PeerIDs to enable communication between different auth nodes
Currently in src/transparency_log/log.rs, there is a method get_authorized_nodes() that returns a Vec<TransparencyLog>. Perhaps we need another similar method that returns a list of PeerIDs to enable communication between different auth nodes.

fishseabowl:
## Description

Fixes pyrsia/pyrsia#1523

This PR refactors `get_authorized_nodes()` method return value and related methods.
```rust
pub fn get_authorized_nodes(&self) -> Result<Vec<PeerId>, TransparencyLogError>
```
## PR Checklist

### Code Contributions

- [ ] I've run the [integrations tests](https://github.com/pyrsia/pyrsia-integration-tests#how-to-set-up-and-run-the-tests)

### DIFF
--- a/src/transparency_log/log.rs
+++ b/src/transparency_log/log.rs
@@ -16,6 +16,7 @@
 
 use crate::artifact_service::model::PackageType;
 use crate::blockchain_service::event::BlockchainEventClient;
+use libp2p::core::ParseError;
 use libp2p::PeerId;
 use log::{debug, error};
 use pyrsia_blockchain_network::error::BlockchainError;
@@ -56,8 +57,8 @@ pub enum TransparencyLogError {
         invalid_hash: String,
         actual_hash: String,
     },
-    #[error("Node with node ID {node_id} already exists in transparency log")]
-    InvalidNodePeerIDFormat { node_id: String },
+    #[error("Invalid peerId format: {0}")]
+    InvalidNodePeerIDFormat(#[from] ParseError),
     #[error("Invalid operation for ID {id}: {invalid_operation}")]
     InvalidOperation {
         id: String,
@@ -288,7 +289,7 @@ impl TransparencyLogService {
         Ok(self
             .find_added_nodes()?
             .iter()
-            .map(|node| PeerId::from_str(&node.node_id).unwrap())
+            .flat_map(|node| PeerId::from_str(&node.node_id))
             .collect::<Vec<PeerId>>())
     }
 
@@ -306,7 +307,7 @@ impl TransparencyLogService {
             }
             Some(nodes) => {
                 for node in nodes {
-                    if node.eq(&PeerId::from_str(peer_id).unwrap()) {
+                    if node.eq(&PeerId::from_str(peer_id)?) {
                         return Err(TransparencyLogError::NodeAlreadyExists {
                             node_id: peer_id.to_owned(),
                         });
--- a/src/transparency_log/log.rs
+++ b/src/transparency_log/log.rs
@@ -283,8 +283,7 @@ impl TransparencyLogService {
         }
     }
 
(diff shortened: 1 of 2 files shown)


dgarus:
We could make a query to the database to get 1 record.
For example, like this:
```
"SELECT * FROM TRANSPARENCYLOG WHERE operation = '(",
            &Operation::AddNode.to_string(),
            "' OR operation = '",
            &Operation::RemoveNode.to_string(),
            "') and node_id = '" + node_id + "' order by timestamp limit 1;"
```
So, it would much faster than retrieving all records related to node operations.

dgarus:
> > So, it would much faster than retrieving all records related to node operations.
> 
> Since the log db table is based on the ledger, which is immutable, it is necessary to find all AddNode and RemoveNode operations. You can find more context here #817 We assume the Add/Remove Node operations will be relatively limited in volume, but should this lead to performance issues, we can optimize by storing a 'current state' in the table, but this is definitely not necessary right now.

At least, we shouldn't retrieve records related to another nodes.
If the last record would be `AddNode` operation, isn't it sufficient 
condition that a node with according `id` already exists?

dgarus (dismissed):
lgtm

src/transparency_log/log.rs
```diff
@@ -964,7 +971,12 @@ mod tests {
         assert!(result_read.is_ok());
         let vec = result_read.unwrap();
         assert_eq!(vec.len(), 1);
-        assert_eq!(vec.get(0).unwrap().node_id, transparency_log.node_id);
+        assert_eq!(
```
tiainen, line 974: When one of the arguments in `assert_eq` equals to `true`, it can be simplified by using `assert` instead:
```
        assert!(vec.get(0).unwrap().eq(...));
```

src/transparency_log/log.rs
```diff
@@ -984,7 +996,10 @@ mod tests {
         assert!(result_read.is_ok());
         let vec = result_read.unwrap();
         assert_eq!(vec.len(), 1);
-        assert_eq!(vec.get(0).unwrap().node_id, node_id);
+        assert_eq!(
```
tiainen, line 999: Simplify to `assert`.

src/transparency_log/log.rs
```diff
@@ -1013,7 +1028,12 @@ mod tests {
         assert!(result_read.is_ok());
         let vec = result_read.unwrap();
         assert_eq!(vec.len(), 1);
-        assert_eq!(vec.get(0).unwrap().node_id, second_node_id);
+        assert_eq!(
```
tiainen, line 1031: Simplify to `assert`.

src/transparency_log/log.rs
```diff
@@ -282,8 +285,12 @@ impl TransparencyLogService {
 
     /// Gets a list of transparency logs of which the operation is AddNode. Returns an error
     /// when no transparency log could be found.
-    pub fn get_authorized_nodes(&self) -> Result<Vec<TransparencyLog>, TransparencyLogError> {
-        self.find_added_nodes()
+    pub fn get_authorized_nodes(&self) -> Result<Vec<PeerId>, TransparencyLogError> {
+        Ok(self
+            .find_added_nodes()?
+            .iter()
+            .flat_map(|node| PeerId::from_str(&node.node_id))
```
tiainen, line 292: This basically hides the parse error, but I suppose we can live with that, assuming that the record is invalid and should not result in an error to the caller of this method.

tiainen (approved):

dgarus (approved):
lgtm

PR #1538, opened 2023-01-23
TransparencyLog: getting an added node record, if it was removed earlier

### LINKED ISSUE #1534: The `transparency_log.find_added_nodes` method doesn't return a record about added node if it was removed earlier

The reproducer:
```
#[tokio::test]
    async fn test_get_authorized_nodes_add_and_remove_and_add() {
        let tmp_dir = test_util::tests::setup();

        let (log, _) = test_util::tests::create_transparency_log_service(&tmp_dir);

        let first_node_id = "node_id_1";
        let transparency_log1 = new_auth_node_transparency_log(Operation::AddNode, first_node_id);
        let res = log.get_authorized_nodes().unwrap();
        assert!(res.iter().any(|t|t.node_id.eq("node_id_1")));

        let transparency_log2 =
            new_auth_node_transparency_log(Operation::RemoveNode, first_node_id);

        assert!(log.write_transparency_log(&transparency_log2).is_ok());

        let res = log.get_authorized_nodes().unwrap();
        assert_eq!(res.len(), 0);

        let transparency_log1 = new_auth_node_transparency_log(Operation::AddNode, first_node_id);
        assert!(log.write_transparency_log(&transparency_log1).is_ok());

        let res = log.get_authorized_nodes().unwrap();
        assert!(res.iter().any(|t|t.node_id.eq("node_id_1"))); // this assert panics

        test_util::tests::teardown(tmp_dir);
    }
```

### Expected Behavior

The `get_authorized_nodes` must return the "node_id_1" record.

rustc 1.66.1 (90743e729 2023-01-10)
Mac OS
v0.2.3

dgarus:
## Description

Fixes #1534
- improve getting a vector of authentication nodes
- fixed bug of getting an added node record, if it was removed earlier
- added according test

## PR Checklist

### Code Contributions


fishseabowl (commented):
Thanks for the great work you do. I have two questions:
1. It is allowed to add the same auth node repeatedly, or delete the same auth node repeatedly, which is not what we expect
2. It may depend on the timestamp, which may lead to some unexpected results.

dgarus:
Thanks!

> It is allowed to add the same auth node repeatedly, or delete the same auth node repeatedly, which is not what we expect
@fishseabowl 

Yes. This fix is addressed to right gathering of data from database, not saving. Assumed that database contains valid data.


> It may depend on the timestamp, which may lead to some unexpected results.

Obviously, `TransparencyLogService` must have guaranties that `timestamp` for every `AuthorizedNode` with the same `id` greater than the previous.
The similar guaranties need for `AddArtifactRequest` requests.
Otherwise, we get mess in `TransparencyLog`. For example, `remove` operation could be earlier `add` operation and so on. In that case, we cannot say what state of record is actual.

These are important questions, but they are out of scope of this PR.

fishseabowl:
@dgarus Thanks for your comments. Since our system is a distributed system, we may have difficulty keeping consistent time. In my opinion, maybe causal order is better.

dasmanas (approved):
lg

src/transparency_log/log.rs
```diff
@@ -1042,6 +1037,30 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_get_authorized_nodes_add_remove_add() {
+        let tmp_dir = test_util::tests::setup();
+        let (log, _) = test_util::tests::create_transparency_log_service(&tmp_dir);
+
+        let node_id: &str = &PeerId::random().to_string();
+        let tl_add_1 = new_auth_node_transparency_log(Operation::AddNode, node_id);
+        assert!(log.write_transparency_log(&tl_add_1).is_ok());
+
```
fishseabowl, line 1048: I found that adding the same node_id repeatedly is still ok. By design, if an authority node is already in the Pyrsia network and has not been removed, we will not add it again to prevent committing useless blocks. Try adding the following test cases::
```rust
let tl_add_1_1 = new_auth_node_transparency_log(Operation::AddNode, node_id);
        assert!(log.write_transparency_log(&tl_add_1_1).is_err());
```
Thanks

dgarus: Yes, I didn't fix double adding because it looks like should be a separate issue.
In this issue, I fixed only retrieving of transparent log records bug.
But I could fix the double adding issue in this PR too if you think it is ok.

dgarus: @fishseabowl
I created the issue #1581 to address double adding the same node.

fishseabowl (approved):
lgtm

PR #1567, opened 2023-01-26
optimization

### LINKED ISSUE #1535: Explore how to efficiently manage the status of authority nodes
Currently in our implementation we reconstruct the status of authority nodes by serially querying database records. Maybe we should explore how to optimize this process.

Some discussion from PR https://github.com/pyrsia/pyrsia/pull/1532 as follows:

[dgarus](https://github.com/dgarus) commented [2 hours ago](https://github.com/pyrsia/pyrsia/pull/1532#issuecomment-1398716007) • 
```
We could make a query to the database to get 1 record in the verify_node_can_be_added_to_transparency_logs.
For example, like this:

"SELECT * FROM TRANSPARENCYLOG WHERE operation = '(",
            &Operation::AddNode.to_string(),
            "' OR operation = '",
            &Operation::RemoveNode.to_string(),
            "') and node_id = '" + node_id + "' order by timestamp limit 1;"
So, it would be much faster than retrieving all records related to node operations.
```

[erwin1](https://github.com/erwin1) commented [2 hours ago](https://github.com/pyrsia/pyrsia/pull/1532#issuecomment-1398743910)
```
Since the log db table is based on the ledger, which is immutable, it is necessary to find all AddNode and RemoveNode operations. You can find more context here https://github.com/pyrsia/pyrsia/issues/817
We assume the Add/Remove Node operations will be relatively limited in volume, but should this lead to performance issues, we can optimize by storing a 'current state' in the table, but this is definitely not necessary right now.
```

[dgarus](https://github.com/dgarus) commented [39 minutes ago](https://github.com/pyrsia/pyrsia/pull/1532#issuecomment-1398833772)
```
At least, we shouldn't retrieve records related to another nodes.
If the last record would be AddNode operation, isn't it sufficient
condition that a node with according id already exists?
```

_Originally posted by @dgarus in https://github.com/pyrsia/pyrsia/issues/1532#issuecomment-1398833772_

dgarus:
## Description

Fixes #1535

In this way we don't retrieve all node's records, initiate structs and iterate through them.
We just get one record and make decision.

## PR Checklist

### Code Contributions

### DIFF
--- a/src/transparency_log/log.rs
+++ b/src/transparency_log/log.rs
@@ -310,18 +310,21 @@ impl TransparencyLogService {
             "SELECT
               operation
             FROM TRANSPARENCYLOG
-            WHERE operation = 'AddNode' or operation = 'RemoveNode' and node_id = '{}'
-            ORDER BY timestamp DESC limit 1", peer_id
+            WHERE (operation = '{}' or operation = '{}') and node_id = '{}'
+            ORDER BY timestamp DESC limit 1",
+            Operation::AddNode.to_string(),
+            Operation::RemoveNode.to_string(),
+            peer_id
         );
 
-        let res = self.open_db()?.query_row(
-            query.as_str(), [], |row| row.get::<usize, String>(0),
-        );
+        let res = self
+            .open_db()?
+            .query_row(query.as_str(), [], |row| row.get::<usize, String>(0));
 
         if let Ok(val) = res {
             if val.eq("AddNode") {
                 return Err(TransparencyLogError::NodeAlreadyExists {
-                    node_id: peer_id.to_owned()
+                    node_id: peer_id.to_owned(),
                 });
             }
         }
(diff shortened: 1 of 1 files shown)


src/transparency_log/log.rs
```diff
@@ -306,23 +306,26 @@ impl TransparencyLogService {
...
-                    }
-                }
-                // was removed, can be added
-                Ok(())
+        let query = format!(
+            "SELECT
+              operation
+            FROM TRANSPARENCYLOG
+            WHERE operation = 'AddNode' or operation = 'RemoveNode' and node_id = '{}'
+            ORDER BY timestamp DESC limit 1", peer_id
```
dasmanas, line 314: Can you please explain the query here?

fishseabowl (commented):
It looks good to me. Thanks

fishseabowl (dismissed):

src/transparency_log/log.rs
dgarus: > Can you please explain the query here?
```
SELECT
  operation -- we wonder only one field, it is enough to understand what is going on
FROM TRANSPARENCYLOG
WHERE 
	(operation = 'AddNode' or operation = 'RemoveNode') -- selects only node operations (I fixed this condition)
	and node_id = '{}' -- selects only certain node records
ORDER BY timestamp DESC  -- the freshest operation is on top
limit 1 -- we need only one record
```
If we get nothing - node with `node_id` is not existed.
If we get 'RemoveNode' - node was removed.
If we get 'AddNode' - node with `node_id` is already existed.

fishseabowl (approved):
lgtm

dasmanas (approved):
lgtm

PR #1569, opened 2023-01-27
getting only specific parts of the transparency log record

### LINKED ISSUE #1551: Extend `inspect-log` to get only specific parts of the `TransparencyLog`
### Suggestion details

Part of #1399

To output only specific parts of the logs.

Query requests should be executed on the node side.

### Additional details

_No response_

dgarus:
## Description

Fixes #1551

- File `node_api/model/cli` renamed  to `node_api/model/request`
- extended REST API to return only specific parts of the transparency log record
- added according tests
- Pyrsia CLI's inspect-log subcommand supports additional argument `--fields` which can accept the valid column names from transparency log table separated by comma.

## PR Checklist

### Code Contributions


fishseabowl:
I guess in the description, `rename node_ip/model/cli to node_ip/model/request` is probably `rename node_api/model/cli to node_api/model/request`. Can you double check it? thanks

dgarus:
@fishseabowl 

> I guess in the description, `rename node_ip/model/cli to node_ip/model/request` is probably `rename node_api/model/cli to node_api/model/request`. Can you double check it? thanks

I renamed `node_ip/model/cli` to `node_ip/model/request` because structs in this mod are related to requests of REST API and I think old name `cli` can be misunderstanding.

fishseabowl:
@dgarus Thanks for your reply. I guess this is a typo, `node_ip` should be `node_api`.

dgarus:
> @dgarus Thanks for your reply. I guess this is a typo, `node_ip` should be `node_api`.

Oh, yes! It is a typo. I didn't notice it, thanks!

fishseabowl (commented):
I ran a test: `./target/debug/pyrsia inspect-log docker --fields id` and it returns `panic`. 
```
xuefeihan@xuefeih-mac pyrsia % ./target/debug/pyrsia inspect-log docker --fields id
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', pyrsia_cli/src/main.rs:101:63
```

Perhaps `panic` should be avoided. thanks

dasmanas (commented):
We may have a better description instead of "added argument --fields to inspect-log CLI command"
Like Pyrsia CLI's `inspect-log` subcommand will now support additional argument `--fields` which can accept the valid column names from transparency log table separated by comma.

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -195,13 +191,19 @@ pub async fn node_list() {
...
+    arg_fields: Option<String>,
+) {
+    let content_type = ContentType::from(arg_format.as_ref()).unwrap();
+    let fields = parse_arg_fields(arg_fields).unwrap();
 
     let result = node::inspect_docker_transparency_log(RequestDockerLog {
         image: image.to_owned(),
         output_params: Some(TransparencyLogOutputParams {
             format: Some(content_type),
+            content: fields,
```
dasmanas, line 206: why the attribute named as content. Would it be make more meaningful with `fields` or `cols` or `columns`?

src/node_api/handlers/swarm.rs
```diff
@@ -17,112 +17,113 @@
...
-        match self {
-            ContentType::JSON => ContentType::JSON,
-            ContentType::CSV => ContentType::CSV,
+impl Serialize for OutputTransparencyLog<'_> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut s = serializer.serialize_struct("TransparencyLog", self.output_fields.size())?;
+        for fld in &self.output_fields.fields {
```
betarelease, line 48: field instead of fld? easy to understand for me.

src/node_api/handlers/swarm.rs
```diff
@@ -17,112 +17,113 @@
...
+            match fld {
+                TransparencyLogField::Id => s.serialize_field("id", &self.origin.id)?,
```
betarelease, line 50: I looked for a better way to do this so that it does not require so much of similar looking code. But it requires building a macro which is not easy to understand. Let's keep it this way.

betarelease (dismissed):
Made minor comments but otherwise looks functional.

dgarus:
@fishseabowl 

> I ran a test: ./target/debug/pyrsia inspect-log docker --fields id and it returns panic.

Oh, You caught a bug from `main` brunch. 
The reason is `image` attribute is omitted. 
I fixed this bug.
We have to have tests for CLI!

src/node_api/handlers/swarm.rs
dgarus: Ok, good point. I think when we will have more places required code like this we could create a macro.

pyrsia_cli/src/cli/handlers.rs
dgarus: Why not `fields` - because here already has `TransparencyLogField` and a collection of this enum values is `fields` (or can be  `columns`).
```
pub struct Content {
    pub fields: Vec<TransparencyLogField>,
}
```
To separate these entities, I decided to use the name `content`.
Otherwise, we would get something like this
`fields.fields` rather than `content.fields`.

dgarus:
@fishseabowl @betarelease @dasmanas 
I fixed comments, please have a look.
Thanks!

pyrsia_cli/src/cli/parser.rs
```diff
@@ -81,19 +81,23 @@ pub fn cli_parser() -> ArgMatches {
                         .about("Show transparency logs for a Docker image")
                         .arg_required_else_help(true)
                         .args(&[
-                            arg!(--image <IMAGE> "The docker image (e.g. alpine:3.15.3 or alpine@sha256:1e014f84205d569a5cc3be4e108ca614055f7e21d11928946113ab3f36054801"),
+                            arg!(--image <IMAGE> "The docker image (e.g. alpine:3.15.3 or alpine@sha256:1e014f84205d569a5cc3be4e108ca614055f7e21d11928946113ab3f36054801")
+                                .required(true),
                             arg!(--format <FORMAT> "The output format")
                                 .value_parser(["json", "csv"])
                                 .default_value("json"),
+                            arg!(--fields <FIELDS> "The list of output fields separated by comma ','. Available fields: id, package_type, package_specific_id, num_artifacts, package_specific_artifact_id, artifact_hash, source_hash, artifact_id, source_id, timestamp, operation, node_id, node_public_key"),
```
dasmanas, line 89: Ideally the field name shouldn't be hardcoded here. This increases debt. Either get it from meta of the table or (if to much work) leave it stating that shouldn't be outside of the field name of Transparency Log table.

pyrsia_cli/src/cli/parser.rs
```diff
@@ -81,19 +81,23 @@ pub fn cli_parser() -> ArgMatches {
...
                         .about("Show transparency logs for a maven artifact")
                         .arg_required_else_help(true)
                         .args(&[
-                            arg!(--gav <GAV> "The maven GAV (e.g. org.myorg:my-artifact:1.1.0)"),
+                            arg!(--gav <GAV> "The maven GAV (e.g. org.myorg:my-artifact:1.1.0)")
+                                .required(true),
                             arg!(--format <FORMAT> "The output format")
                                 .value_parser(["json", "csv"])
                                 .default_value("json"),
+                            arg!(--fields <FIELDS> "The list of output fields separated by comma ','. Available fields: id, package_type, package_specific_id, num_artifacts, package_specific_artifact_id, artifact_hash, source_hash, artifact_id, source_id, timestamp, operation, node_id, node_public_key"),
```
dasmanas, line 100: Ideally the field name shouldn't be hardcoded here. This increases debt. Either get it from meta of the table or (if to much work) leave it stating that shouldn't be outside of the field name of Transparency Log table.

dgarus: Good point, thanks! I formed `fields` help string based on default `Content` struct.
I don't think that we should bind output fields to log table's fields. We could output fewer fields than in the table.

pyrsia_cli/src/cli/handlers.rs
```diff
@@ -235,6 +243,22 @@ pub async fn inspect_maven_transparency_log(gav: &str, format: Option<String>) {
...
+        let fields: Vec<TransparencyLogField> = val
+            .split(',')
+            .map(|s| s.trim())
+            .map(|s| s.parse::<TransparencyLogField>().unwrap())
+            .collect();
+
+        Some(Content { fields })
+    } else {
+        None
+    })
```
ihcomega56, line 259: Can this part be written like this...? 

```rust
    Ok(arg_fields.map(|f| {
        let transparencyLogFields = f
            .split(',')
            .map(|s| s.trim())
            .map(|s| s.parse::<TransparencyLogField>().unwrap())
            .collect::<Vec<TransparencyLogField>>();
        Content {
            fields: transparencyLogFields,
        }
    }))
```

dgarus: Yes, great. 
Thanks!

ihcomega56 (approved):
Now it looks good to me 👍

efrisch (approved):
One small question about implementation detail, but not a blocker.

src/node_api/handlers/swarm.rs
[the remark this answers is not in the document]
dgarus: This is the main feature of this PR!
Here, we add to response only fields which a user ordered.
Requested fields are contained in`output_fields.fields`.

PR #1584, opened 2023-02-03
Repeatedly adding the same node

### LINKED ISSUE #1581: Repeatedly adding the same node

Adding the same node_id repeatedly is still ok.
Reproducer can be found [here](https://github.com/pyrsia/pyrsia/pull/1538#discussion_r1094935808).

### Expected Behavior

 By design, if an authority node is already in the Pyrsia network and has not been removed, we will not add it again to prevent committing useless blocks.

rustc 1.66.1 (90743e729 2023-01-10)
Mac OS
v0.2.3

dgarus:
## Description

Fixes #1581 

- added node checks for `TransparencyLogService::remove_authorized_node` method
- made `TransparencyLogService::write_transparency_log` method private
- extended `TransparencyLogService::add_artifact` to insert a log record to the database
- added according tests

## PR Checklist

### Code Contributions

### DIFF
--- a/src/artifact_service/service.rs
+++ b/src/artifact_service/service.rs
@@ -153,13 +153,11 @@ impl ArtifactService {
                 .transparency_log_service
                 .add_artifact(add_artifact_request)
                 .await?;
+
             info!(
                 "Transparency Log for build with ID {} successfully created.",
                 build_id
             );
-
-            self.transparency_log_service
-                .write_transparency_log(&add_artifact_transparency_log)?;
 
             self.put_artifact_from_build_result(
                 &artifact.artifact_location,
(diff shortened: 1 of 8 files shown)


src/artifact_service/service.rs
```diff
@@ -217,13 +215,9 @@ impl ArtifactService {
...
-            if let Err(TransparencyLogError::LogNotFound { .. }) = self
-                .transparency_log_service
-                .find_transparency_log(&transparency_log.id)
-            {
-                self.transparency_log_service
-                    .write_transparency_log(&transparency_log)?;
-            }
+            self.transparency_log_service
+                .write_if_not_exists(&transparency_log)
+                .await?;
```
fishseabowl, line 220: It looks good to me. But maybe this wrapper is unnecessary.

src/transparency_log/log.rs
```diff
@@ -226,9 +239,24 @@ impl TransparencyLogService {
             .add_block(payload.into_bytes())
             .await?;
 
+        self.write_transparency_log(&transparency_log)?;
+
```
fishseabowl, line 243: Could you explain the advantages of this modification in detail? Thanks

dgarus: 1. After every calling `add_artifact` follows `write_transparency_log`. So, the first reason is decreasing code duplication because we don't have reason to repeat this code block in different places.
2. If we want to control what we write to the database, we should encapsulate `write_transparency_log` otherwise we have no context for checks. Hence, we should use this method only inside TransparencyLogService's high-level methods like 'add_artifact', 'remove_authorized_node' and so on.

src/artifact_service/service.rs
dgarus: The main aim is encapsulation `write_transparency_log`.

fishseabowl (approved):
lgtm

dasmanas (approved):

PR #1600, opened 2023-02-08
Moved gossipsub block broadcast to after all of the artifact is created.

### LINKED ISSUE #1246: Artifact p2p provide should happen before new block broadcast

efrisch:
## Description

Fixes pyrsia/pyrsia#1246

This PR delays broadcasting new blocks until after the artifact(s) are all added by making the broadcast a separate function, it takes a vec of payloads, and that happens after the loop adding all of the artifact component files.

## Screenshots (optional)
Integration Tests Run

## PR Checklist

### Code Contributions

### DIFF
--- a/src/artifact_service/service.rs
+++ b/src/artifact_service/service.rs
@@ -135,6 +135,7 @@ impl ArtifactService {
             build_id, build_result.package_type, package_specific_id
         );
 
+        let mut payloads: Vec<String> = Vec::new();
         for artifact in build_result.artifacts.iter() {
             let add_artifact_request = AddArtifactRequest {
                 package_type: build_result.package_type,
@@ -149,11 +150,13 @@ impl ArtifactService {
                 add_artifact_request
             );
 
-            let add_artifact_transparency_log = self
+            let add_artifact_transparency_tuple = self
                 .transparency_log_service
                 .add_artifact(add_artifact_request)
                 .await?;
 
+            let add_artifact_transparency_log = add_artifact_transparency_tuple.0;
+            payloads.push(add_artifact_transparency_tuple.1);
             info!(
                 "Transparency Log for build with ID {} successfully created.",
                 build_id
@@ -170,6 +173,9 @@ impl ArtifactService {
                 .await?;
         }
 
+        self.transparency_log_service
+            .broadcast_artifacts(payloads)
+            .await?;
         Ok(())
     }
 
@@ -439,7 +445,7 @@ mod tests {
         let package_type = PackageType::Docker;
         let package_specific_id = "package_specific_id";
         let package_specific_artifact_id = "package_specific_artifact_id";
-        let transparency_log = artifact_service
+        let transparency_log_tuple = artifact_service
             .transparency_log_service
             .add_artifact(AddArtifactRequest {
                 package_type,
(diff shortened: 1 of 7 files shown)


dasmanas (dismissed):
lgtm 👌

src/transparency_log/log.rs
```diff
@@ -231,17 +231,30 @@ impl TransparencyLogService {
     pub async fn add_artifact(
         &self,
         add_artifact_request: AddArtifactRequest,
-    ) -> Result<TransparencyLog, TransparencyLogError> {
+    ) -> Result<(TransparencyLog, String), TransparencyLogError> {
         let transparency_log = TransparencyLog::from(add_artifact_request);
 
         let payload = serde_json::to_string(&transparency_log)?;
+        /*
```
karolh2000, line 238: probably it'd be better to remove this dead code? any particular reason why it wasn't removed?

karolh2000 (dismissed):

src/transparency_log/log.rs
efrisch: Removed now.

karolh2000 (approved):

src/artifact_service/service.rs
```diff
@@ -149,11 +150,13 @@ impl ArtifactService {
...
             );
 
-            let add_artifact_transparency_log = self
+            let add_artifact_transparency_tuple = self
                 .transparency_log_service
                 .add_artifact(add_artifact_request)
                 .await?;
 
+            let add_artifact_transparency_log = add_artifact_transparency_tuple.0;
+            payloads.push(add_artifact_transparency_tuple.1);
```
fishseabowl, line 159: Looks good. But `add_artifact_transparency_tuple.0` and `add_artifact_transparency_tuple.1` seem confusing to me. Maybe structure is better.

dasmanas (approved):
lgtm

fishseabowl (commented):
From your code, the build request workflow looks like this: when a build request is triggered, some artifacts are ready, then these artifacts are added to the transparency_log(`transparency_log_service.add_artifact`), next the swarm is notified that the node is a provider (`p2p_client.provide`), and finally these artifacts are added to the blockchain and broadcast (`transparency_log_service.broadcast_artifacts`). 
I'm not sure if I understand your code correctly. Could you please provide details on this workflow? Thanks

Issue #1454, opened 2022-12-22
Adding Test Cases for CLIs


dasmanas:
### Suggestion details

Adding test cases for CLIs.

As per my exploration here is the details.

```
[dev-dependencies]
assert_cmd = "2.0.5"
predicates = "2.1.1"
```

Sample test code. The test code runs the binary from the debug directory.
```
#[cfg(test)]
#[cfg(not(tarpaulin_include))]
mod tests {
    use crate::CONF_FILE_PATH_MSG_STARTER;
    use assert_cmd::Command;
    use predicates::prelude::*;

    #[tokio::test(flavor = "multi_thread")]
    async fn test_config_show_file_path_info_availability() {
        let mut cmd = Command::cargo_bin("pyrsia").unwrap();
        cmd.arg("config")
            .arg("--show")
            .assert()
            .success()
            .stdout(predicate::str::contains(CONF_FILE_PATH_MSG_STARTER));
    }
}
```

The code expects the binary will be present under `target/debug` directory. But tarpaulin is not going to build the binary under debug. Therefore a cargo build has to perform. In subsequent step while calling cargo tarpaulin, we need to preserve previously build debug dir. To achieve the same, add `--skip-clean` flag to `cargo tarpaulin`.

Please refer commits from 
- https://github.com/pyrsia/pyrsia/pull/1245/commits/0c6903b6269d8133608ac930e4ddab31daebef7e
- https://github.com/pyrsia/pyrsia/pull/1245/commits/c870282ed43c0fbda7bbe04bd3fd51b4e54799a9

_Disclaimer: Even if you will be able to run unit test but the coverage is not going to be included in code coverage report._

PR #1644, opened 2023-02-21
Prototype of integration testing based on `testcontainers`

dgarus:
Referred to #1454

`pre-commit.sh` builds the `pyrsia/test_node` image, runs tests and removes the image after tests.

The flag `ignore-it` allows ignoring integration tests.

Evolution:
 - Create Dockerfile for tests purposes as test resource.
 - Prepare SQLite db file and copy it into container to test node responses (maybe, is there a better idea?)
 - To test p2p communications between nodes, we should have a test discovery service. Do we have a test discovery service?

### DIFF
--- a/.github/workflows/.cargo/audit.toml
+++ b/.github/workflows/.cargo/audit.toml
@@ -2,4 +2,4 @@
 # A good example is at https://docs.rs/crate/cargo-audit/0.10.0/source/audit.toml.example
 
 [advisories]
-ignore = ["RUSTSEC-2022-0040"] # advisory IDs to ignore
+ignore = ["RUSTSEC-2022-0040", "RUSTSEC-2020-0071"] # advisory IDs to ignore
--- a/.github/workflows/.cargo/audit.toml
+++ b/.github/workflows/.cargo/audit.toml
@@ -2,4 +2,7 @@
 # A good example is at https://docs.rs/crate/cargo-audit/0.10.0/source/audit.toml.example
 
 [advisories]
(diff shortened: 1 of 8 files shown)


betarelease:
To get this reviewed, please convert this to `ready for review`

dgarus:
> To get this reviewed, please convert this to `ready for review`

Ok

dgarus:
image

https://rustsec.org/advisories/RUSTSEC-2020-0071

Can we use `testcontainer` with this vulnerability if it is used only in a test scope?

fishseabowl:
> image
> 
> https://rustsec.org/advisories/RUSTSEC-2020-0071
> 
> Can we use `testcontainer` with this vulnerability if it is used only in a test scope?

In my opinion it is better to avoid any vulnerability error. Maybe you can try the latest version `time 0.3.19`.

fishseabowl (commented):
Some checks failed. It is better to fix it. Thanks

dgarus:
> > image
> > https://rustsec.org/advisories/RUSTSEC-2020-0071
> > Can we use `testcontainer` with this vulnerability if it is used only in a test scope?
> 
> In my opinion it is better to avoid any vulnerability error. Maybe you can try the latest version `time 0.3.19`.

I can't change this

https://github.com/testcontainers/testcontainers-rs/issues/389
```
RUSTSEC-2020-0071 Impact
Unix-like operating systems may segfault due to dereferencing a dangling pointer in specific circumstances. 
```
It doesn't look critical for tests, isn't it?

